PR #27, opened 2017-03-21
Fixes #24 - All buttons converted to use <a> tags and hrefs

### LINKED ISSUE #24: Change all buttons to be <a> elements
Currently most buttons are `<div>` elements.

They should be `<a>` elements.

JS will need to be updated - instead of `.onclick = function {}`, it will need to be `.href = ""`.

Some classes, like no-underline, will need to be added to the <a> elements to keep them looking the same.

Joe-Brady:
Issue #24 fixed by this PR.

All buttons were previously `<div>` elements, now are `<a>`.

This means people can open links in new tabs if they wish, there is less JS all round, a more standardised way of linking to other pages, etc.

Also changed links to use .html at the end, instead of relying on GitHub to add this silently.
This is because local development was frustratingly slow, with .html needing to be added every time a link was clicked.

We can remove the .htmls again later, if desired, once we have a way to solve the local development issue.


Joe-Brady:
Yes - I was loathe to use it, but in this case I was struggling to find the 'right way' to do it. Will keep it to a minimum.

PR #34, opened 2017-03-22
Add menu

Joe-Brady:
A new menu bar for every page. 
Suitable for all screen sizes.
Slides open and closed.

@tellison I have agreed with Martijn to remove the GetOpenJDK wording for now until we have a more permanent solution, because it does not fit on mobile screens, and is a little confusing for visitors.


Joe-Brady:
@karianna Yes I think that's a good idea. Any suggestions? I've only used Handlebars/partials before, and that was ok, but there must be better solutions

karianna:
I think Mustuache / Handlebars are common, I'm told that https://github.com/pugjs/pug (formerly Jade) is also a contender these days.  But happy to go with your choice!

PR #50, opened 2017-04-05
New nightly table, Linux s390x nightly builds, and more

Joe-Brady:
STAGING PREVIEW: http://staging.adoptopenjdk.net/nightly.html

**New nightly page table:** each platform now appears on a new table row, allowing for improved searching, and partially setting the stage for more platforms to be added in the future.
> Bonus feature: because the new table is narrower, the "Scroll horizontally to see all information" text will only appear if the user's window is narrow enough to activate the table's horizontal scrollbar.

**s390x nightly builds now visible on website:** Using the new nightly table JS, it was an easier job to add the new s390x builds. There is, however, still more code that needs to be changed to make it truly future-proof.
> Note: this is for nightly builds only, not Releases or Archive yet!

**404 links removed:** The releases, archive, and nightly pages now have temporarily removed links and sections for: Build Number, Changelog, Commit Ref. Will be re-added once this is available in the API.

**Naming updated:** Linux is now called Linux x86-64, and we also have Linux s390x on the nightly page only.


src/js/releases.js
```diff
@@ -94,7 +96,7 @@ function populateLatest() {
 
         // set the download links for this release...
 
-        if(a.indexOf("LINUX") >= 0) { // if the binary name contains 'LINUX':
+        if(a.indexOf("X64_LINUX") >= 0) { // if the binary name contains 'LINUX':
```
tellison: This check may need tweaking again in the near future when we fix the binaries to have the correct CPU architecture identifiers.

src/js/nightly.js
```diff
@@ -64,89 +63,99 @@ function populateNightly() {
...
+
+            // TODO: this should be abstracted - a global function should be able to search the filename for different content and return the correct platform name.
+            if(a.indexOf("S390X_LINUX") >= 0) {
+              document.getElementById("platform-block"+tableRowCounter).innerHTML = "Linux s390x";
+            } else if(a.indexOf("X64_LINUX") >= 0) {
+              document.getElementById("platform-block"+tableRowCounter).innerHTML = "Linux x86-64";
+            } else if (a.indexOf("WIN") >= 0) {
+              document.getElementById("platform-block"+tableRowCounter).innerHTML = "Windows";
+            } else if (a.indexOf("MAC") >= 0) {
+              document.getElementById("platform-block"+tableRowCounter).innerHTML = "macOS";
```
tellison: Windows and macOS should also have CPU architecture identifier (both will be x86-64 presently) to be consistent.

Joe-Brady: No problem - I also have a solution in mind (that I will implement in 2 weeks' time) that works by having an array of JS objects.

Each object will represent an 'accepted platform', and will contain information about that platform, such as the 'official name' (including x86-64 etc), the associated file extension (.zip, .tar.gz, etc), and any other information we want to include.

This will ultimately be the one single place where a change is required, in order to add a new platform to the website. The rest of the JS, HTML etc. will be abstracted to deal with as many objects (platforms) as you want to put into the array.

Joe-Brady: Will add the x86-64 identifier across the site now

Joe-Brady:
UPDATED ON STAGING: http://staging.adoptopenjdk.net/nightly.html

(also with bonus feature for 'entry level users' - click on the home page's main download button, to be visually guided towards the Installation page)

Issue #35, opened 2017-03-22
Re-structure of website once more platforms come onboard


karianna:
https://nodejs.org/en/download/ as an example via @gdams

Joe-Brady:
The latest release page, the archive page, and the nightly page are all currently set up to handle Windows as well as Linux and Mac.

The latter will need to be manually changed (see #31 ) following the first Windows build.

The other two pages will automatically show the Windows build.

However, beyond these three operating systems, while the visual appearance of the website will not require too much of a redesign, the JS change is a major undertaking, as I originally designed the website with the expectation that there would only be 3 operating systems.

With these points in mind, what would you like this issue to represent? I.e. the addition of Windows is already covered.

Joe-Brady:
Regarding the reference of Node.js from @gdams, this was already considered at the beginning of the project.

As soon as there is a second release in the 'releases' repo, you will see how the Archive page provides a more condensed version of the 'latest release' information for each release. It's a shame you can't already see it without us creating a 'demo' release!

gdams:
I think that @karianna's concern is more that we are already planning to have linux-x64, linux-s390x, linux-ppc64, win-x64 and macos-x64 releases

gdams:
with so many different architectures, the current 'logo' based system may not work very well

karianna:
@gdams is correct (and also Solaris at some stage).  I'd argue this is a backlog issue until it happens but wanted to give a heads up :-).

Joe-Brady:
This is true - although that kind of visual change is minor compared to the rather enormous task of re-writing the JS to be able to deal with whatever platforms we throw at it.

gdams:
also would be nice to add arm to this

Joe-Brady:
Exactly - there are so many 'it would be nice to add...' that the JS needs to be almost re-written from scratch. Currently, the terms 'linux', 'windows' and 'mac' are hard-coded absolutely everywhere

Joe-Brady:
Just to give you an idea of the scale of this task to incorporate anything new that's not Windows

gdams:
we probably need a re-write of the js at some point as we need to consider architectures as well as multiple file extensions

tellison:
The current site design is great, and provides a good first impression showing what is available in the early days; but I agree that restructuring the js to allow for a growing number of platforms should be considered.  Hopefully that is something you can introduce over time without breaking the existing site.

tellison:
[...]

PR #56, opened 2017-04-07
Nightly page abstracted to accept any defined platform

Joe-Brady:
Work in relation to #35
CURRENTLY ON STAGING: http://staging.adoptopenjdk.net/nightly.html

Completed the first step towards full abstraction in the JS to accept any defined platform.

`nightly.js` will now accept ALL platforms, as long as they are defined in the 'platforms' array at the top of `global.js`.

E.g. I have now simply added this entry to the array:
```
{
    officialName: "Linux ppc64le",
    searchableName: "PPC64LE_LINUX",
    logo: "linux.png",
    fileExtension: ".tar.gz"
  },
```

Meaning that the PPC builds will now appear on the Nightly page.
We can also optionally remove any platform from the Nightly page by removing it from the array.

**Also of note:** I have added an entry for Linux arm builds. Currently, arm builds will only be a 'match' (and therefore be displayed on the Nightly page) if the filename includes "arm_Linux" (any case).

My next step is to roll out usage of this new array-based system to the rest of the website. Will do this when there's time.


Joe-Brady:
Travis has failed based on some unused vars - they are not actually unused, so I'll add exceptions for these...

tellison:
Had a read through and LGTM.

dist/js/app.js
```diff
@@ -1,7 +1,85 @@
...
+
+// gets the FILE EXTENSION when you pass in 'searchableName'
+function getFileExt(searchableName) {
+  return (lookup[searchableName].fileExtension);
+}
+
+// set path to logos
+var logoPath = "/dist/assets/";
+console.log(logoPath); // TODO - REMOVE THIS LINE ONCE LOGOPATH IS USED
+
```
karianna: Do you mean when `logoPath` isn't hard coded?

dist/js/app.js
```diff
@@ -136,7 +210,9 @@ function setDownloadSection() {
       }
       else { // if there IS a matching binary for the user's OS:
         var fullOSName = OS; // defaults this variable to be the detected OS name
-        if(OS == "Win") {
+        if(OS == "Linux") {
+          fullOSName = "Linux x86-64"; // add 'x86-64'
+        } else if(OS == "Win") {
```
karianna: Is it always Linux x86-64?  In general I wonder if this functionality can be replaced with a well known borwser and O/S detection lib.

dist/js/app.js
```diff
@@ -255,65 +333,63 @@ function populateNightly() {
...
-            if(a.indexOf("S390X_LINUX") >= 0) {
-              document.getElementById("platform-block"+tableRowCounter).innerHTML = "Linux s390x";
-            } else if(a.indexOf("X64_LINUX") >= 0) {
-              document.getElementById("platform-block"+tableRowCounter).innerHTML = "Linux x86-64";
-            } else if (a.indexOf("WIN") >= 0) {
-              document.getElementById("platform-block"+tableRowCounter).innerHTML = "Windows x86-64";
-            } else if (a.indexOf("MAC") >= 0) {
-              document.getElementById("platform-block"+tableRowCounter).innerHTML = "macOS x86-64";
+          var a = nameOfFile.toUpperCase(); // make the name of the file uppercase
+          var thisPlatform = getSearchableName(a); // get the searchableName, e.g. MAC or X64_LINUX.
```
karianna: A var named`a`?  :-)

dist/js/app.js
```diff
@@ -255,65 +333,63 @@ function populateNightly() {
...
+              document.getElementById("nightly-checksum"+tableRowCounter).href = (assetArray[assetCounter2].browser_download_url).replace(thisFileExtension, ".sha256.txt"); // set the checksum link (relies on the checksum having the same name as the binary, but .sha256.txt extension)
+              var link = (assetArray[assetCounter2].browser_download_url);
+              dlButton.href = link; // set the download link
+
+              // show the new row
+              var trElement = document.getElementById(tableRowCounter);
+              trElement.className += " animated fadeIn"; // add the fade animation
+              trElement.className = trElement.className.replace( /(?:^|\s)hide(?!\S)/g , '' ); // remove the 'hide' class immediately afterwards
+
+              tableRowCounter++;
```
karianna: This function could be split up a little more (Single Responsibility Principle) into separate functions.

.eslintignore
```diff
@@ -1,2 +1,2 @@
-dist/js/app.js
 dist/js/app.min.js
+src/js
```
Joe-Brady: Yes, because all of our source js is built into app.js - so if app.js is being linted, then everything is being linted.

It's better this way around, because there were linting issues cropping up for the individual js files in /src/js - e.g. functions that exist in global.js, being called by nightly.js, were being flagged as issues. So I could escape all of them, but it seemed much better to just lint app.js

dist/js/app.js
Joe-Brady: Yes, sorry for the poor explanation here!

dist/js/app.js
Joe-Brady: You're right, this is a temporary solution for the homepage, because the homepage is still incapable of providing the other types of Linux downloads.

Could you please open an issue for this? If there are OS detection libraries out there, it would make sense to use one.

dist/js/app.js
Joe-Brady: Yeah it could be a better name!

PR #58, opened 2017-04-07
Nightly search bar improvement

### LINKED ISSUE #57: Improve Nightly page search
Currently, the search bar does not allow for 'mix and match' searches, e.g. 'mac june' or 'jdk8u152-b01 linux'. Results will only be returned for exact matches.

'Mix and match' search functionality appears to be difficult to do with JS alone, but can be fairly easily done with jQuery, which I will add to that one page only, to avoid loading it on the rest of the website.

Joe-Brady:
Fixes #57 
ON STAGING: http://staging.adoptopenjdk.net/nightly.html

The search bar now allows for much more flexible, 'mix-and-match' searching. Examples of searches that would not have worked before, but now do:

"x86 march linux"
"s390 6th apr"
"20170704 mac"
"72 mb 2017 march"


nightly.html
```diff
@@ -106,7 +109,7 @@ <h3 id="scroll-text" class="hide" style="margin: 2rem 0 0 0;">Scroll horizontall
 </footer>
 
 <script src="dist/js/app.min.js"></script>
-<script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.18.0/moment.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.18.0/moment.js"></script><script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
 <script>onNightlyLoad();</script>
```
karianna: newline?

dist/js/app.js
```diff
@@ -255,49 +255,16 @@ function setDownloadSection() {
...
-    }
-  });
-
 }
 
 
 // NIGHTLY PAGE FUNCTIONS
 
 function populateNightly() {
+  const tableHead = document.getElementById("table-head");
```
karianna: This function could do with being split up into functions that represent the discrete bits of work, e.g. The inner HTML bit, the logi search bit etc.

Joe-Brady: Could this be raised as an issue and resolved in a later PR?

I agree that it should be split, but it isn't a new problem being introduced by this PR, so I'd prefer to get this merged and split the function after.

nightly.html
Joe-Brady: This is a result of the current Handlebars setup, where scripts are injected into the footer via a variable.

In this case, I'd already used the two slots that I'd allocated for scripts, so I just put two in the same variable - hence they ended up on the same line.

This could be improved, I agree. Will require a 3rd slot to allow 3 scripts to be injected via Handlebars.

PR #69, opened 2017-04-19
Split on-load functions for maintenance

Joe-Brady:
I have separated out some of the bigger sections of code for each page so that they are in distinct functions of their own (such as `buildArchiveHTML();`).

This meant that some variables (the ones that just declare a name for a HTML element) had to be moved to the global scope. Otherwise would have had to pass a lot of variables into various nested functions.

Also have made a couple of improvements to make the 'loading dots' always accessible under a single variable name (as long as we continue to give the dots element the same ID on each page).


Joe-Brady:
Forgot to mention in the PR description, the first commit removes all of the counter vars, and uses the forEach index as discussed - so that's now done too

karianna:
@Joe-Brady I think this PR actually broke the main site (for me as a Mac user anyhow) - can you check?

Joe-Brady:
@karianna not for me - I'm also a Mac user, absolutely no issues here...

Just in case, before we roll anything back, can you check:
With a different browser
With your usual browser, but without a cache? Either using the dev tools 'no cache' option, or doing a hard refresh - in Chrome on Mac it's Shift-Cmd-R

karianna:
@Joe-Brady Thanks - that's resolved it.

gdams:
@Joe-Brady staging ?

Joe-Brady:
@gdams It was on there at the time of the PR, I forgot to post a link though

PR #72, opened 2017-04-24
Add full OS detection capability and platform abstraction to homepage

### LINKED ISSUE #60: Incorporate a 'std' browser and O/S detection lib so users are presented with the best download

Joe-Brady:
**Important** - before merging this PR, we need to check that the staging server displays the Linux x86-64 download on all Linux-based OSes that we can lay our hands on: http://staging.adoptopenjdk.net/

Fixes #60, and provides the final piece in the puzzle for fixing #35.
 
The homepage can now reliably detect a much longer list of operating systems, thanks to the use of platform.js.

It can also detect 64-bit and 32-bit architectures, which will be useful if you end up doing some 32-bit builds, as I think someone suggested.

The new setup also means that in theory, the homepage could display a download button for any platform. However, in practice, we will still only have Linux x86-64, Windows, and Mac, because any other Linux architectures are going to default to displaying x86-64 anyway, as discussed in Issue #60. I think we may have to stick with this approach, because while it is possible to detect the Debian OS for example, detecting a s390x/ppc64le architecture doesn't seem as easy, and might actually be impossible based on my findings so far.


Joe-Brady:
Bear with me - just changing a few small bits, won't be long.

dist/js/app.js
```diff
@@ -111,16 +131,23 @@ menuClose.onclick = function() {
...
+    /* eslint-disable */
+    var platformFamily = platform.os.family.toUpperCase(); // platform.os.family is dependent on 'platform.js', loaded by index.html (injected in index.handlebars)
+    /* eslint-enable */
+    if(thisPlatformMatchingString.indexOf(platformFamily) >= 0) { // if the detected 'platform family' string appears in the osDetectionString value of a platform...
+      matchedOS = eachPlatform;
+    }
+  });
+
+  // if matchedOS has a value, return it
+  if(matchedOS){ return matchedOS; } else { return null; }
```
karianna: redundant comment

PR #74, opened 2017-04-24
Add download size in MB to home page

### LINKED ISSUE #26: Add download size in MB to home page "Latest release" button
Note: this should only appear if the button is linking to the download directly - not if it is the generic 'no appropriate OS' link.

Joe-Brady:
Fixes #26. On Staging.

Also made the style of that smaller text on the blue button match the equivalent on the Latest Release page.

Also removed parentheses from the blue buttons on Latest Release page, for a slightly improved appearance.

I think we can ignore the change to archive.html - Git seemed to get confused with the last change from Martijn, I didn't actually change that, it's just carried over from his PR.


Joe-Brady:
Just added one more commit

PR #76, opened 2017-04-26
Add social links to footer (Fixes #75)

### LINKED ISSUE #75: Add a row of social media links
Possibly in the footer

Joe-Brady:
Fixes #75 
On [staging](http://staging.adoptopenjdk.net)

Used white logos with transparency to make the icons less obtrusive.
Short fade to full opacity on hover.

Also added a subtle transparency transition on hover for the other icons in the footer.


Joe-Brady:
Added a second commit removing the Having Issues link in the footer - there is a clearer route for getting help via the Support page, and the removal of that link makes the footer look a little cleaner.

PR #109, opened 2017-05-18
Latest page redesign prep

Joe-Brady:
Preparatory changes before the full redesign of the Latest page:

- Removed requirements from Latest page, and from the rest of the code
- Split Archive page's SCSS into its own file
- Changed 'release' to 'build' globally
- Tweaked styling of Archive page download buttons for width consistency
- Switched B&W windows and linux logos for colour.


gdams:
It shows it at the top doesn't it ?

karianna:
It does yes, the `Name: jdk8u152-b04` section, but I don't think that's clear / prominent enough in that case :-)

gdams:
I'm going to land this. We can address Martijn's concerns in the new design

PR #113, opened 2017-05-22
Nightly search bar location, text changed (Fixes #99), & Latest info toggling

### LINKED ISSUE #99: Nightly 'number of builds' improvement
Currently, you can choose to view 10, 20, 50, or 100 nightly builds in the table.
You can also search these builds.

If you have selected '20' builds, and have searched for 'mac', you might expect to see 20 Mac builds. 
However, this is not the case. Instead, the 'mac' search is run **within** the 20 builds, so you might only see approx. 5 results.

Change the JS so the '20', for example, only shows the first 20 that are shown as a result of the search, and not the other way around.

Joe-Brady:


Joe-Brady:
@gdams Just added the toggling as per your request

PR #121, opened 2017-05-30
Add hashes to minified JS and CSS files, add BrowserSync (Fixes #108)

### LINKED ISSUE #108: Resolve Cloudflare cacheing issues
To avoid cacheing preventing new changes from showing up:

**Option 1:**
Cache Control headers etc is a bit of a minefield.  Apache and NginX and other ‘web’ or ‘app’ servers have different ways of configuring them, but they are a std so to speak and CDNs, Browsers etc are supposed to adhere to them.  This is the field of web performance tuning.  Not sure if you’ve got someone in the org who’s an expert, if not then I’d recommend one day setting up a basic site (or a wordpress site with w3total cache plugin if you’re more comfortable with that), get that site fronted by a CDN and start seeing how the different values mean whether the browser just looks locally, goes to the CDN or goes directly to the server.

**Option 2:** 
The CSS and JS files need hashes/unique identifiers in the names every time they are built, and these unique names need to be injected into the HTML. That way, it can’t be cached when the files change, because it looks like a new file.

Once finished, the Cloudflare cache settings can be returned to the normal length of time.

Joe-Brady:

Hashes are now automatically added to the filename of minified CSS and JS.
Links to these hashed files are then automatically injected into every HTML file, so there is no action needed on the part of the developer.

Browser-sync is also added, for instant automatic browser refreshes whenever a file in `/src` is saved.
Browser-sync also allows for easy testing on mobile devices, as explained in the new "PRO TIP" on Contributing.md.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,2 @@
-dist/js/app.min.js
+dist/js/app.min*.js
 src/js
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@
     npm install
     ```
 
-3. Start the auto-build scripts, and leave them running
+3. Start the auto-build scripts and BrowserSync (opens a new `localhost:3000` browser tab). Leave this process running during development.
 
     ```
     npm start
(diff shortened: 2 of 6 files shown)


Joe-Brady:
@karianna I have added an issue for this addition - going to write about the entire Gulpfile though, which will include the hashing and injection.

PR #136, opened 2017-06-06
Add Archive pagination, comment out platform filter (Fixes #126)

### LINKED ISSUE #126: Archive page organisation
Show only the 5 latest releases on archive page.
Work out how to do 'pages' of releases (with next, previous etc)
Also a 'view all' button?

Joe-Brady:
[STAGING](https://staging.adoptopenjdk.net/archive.html) (currently set to max. 5 archive entries per page, so the pagination addition is hidden as there is only 1 page)

This is what it will look like once there are more than 5 archive entries:

screen shot 2017-06-06 at 15 16 02

### DIFF
--- a/src/handlebars/archive.handlebars
+++ b/src/handlebars/archive.handlebars
@@ -1,4 +1,4 @@
-{{> header title='Archive | ' }}
+{{> header title='Archive | ' style1='<link rel="stylesheet" href="http://pagination.js.org/dist/2.0.7/pagination.css">' }}
 
 <main class="grey-bg">
 
@@ -27,6 +27,7 @@
     </div>
 
     <div id="archive-list" class="hide">
+      <div id="pagination-container"></div>
       <table class='archive-container'>
         <tbody id='archive-table-body'></tbody>
       </table>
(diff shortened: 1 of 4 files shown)


src/handlebars/archive.handlebars
```diff
@@ -1,4 +1,4 @@
-{{> header title='Archive | ' }}
+{{> header title='Archive | ' style1='<link rel="stylesheet" href="http://pagination.js.org/dist/2.0.7/pagination.css">' }}
```
karianna: I'm wondering if we need to start thinking about requireJS

src/handlebars/archive.handlebars
```diff
@@ -19,14 +19,17 @@
 
     <div id="loading"><img src="dist/assets/loading_dots.gif" width="40" height="40"></div>
 
+    <!-- SEE THE COMMENTED-OUT FUNCTIONS IN ARCHIVE.JS
```
karianna: Then can we kill the commented out code?  OK, I see what you're doing, I think a better way to handle this is to save that functionality off in a branch or as a patch or even just have a wiki entry stating that if we need it back go look at commit X.  It's usually a bit frowned upon to clutter up the source with lots of commented out code..

src/handlebars/archive.handlebars
Joe-Brady: I've touched on requireJS before, but only for importing and exporting individual parts of JS files into one another... what would be the improvement it brings in this situation? This highlighted code is just injecting the links into the HTML files via the Handlebars templating.

More than happy to add RequireJS if it's necessary

src/handlebars/archive.handlebars
Joe-Brady: Good point... doing that now.

src/handlebars/archive.handlebars
karianna: hmm maybe it's overkill - I'm just concerned we'll lag in the libs that we're pulling in.

PR #144, opened 2017-06-16
Handlebars rewrite, JCK ticks ready, localize fixup, scroll arrows, social image

### LINKED ISSUE #140: Add 'tick' icon, and logic to display it for verified builds

### LINKED ISSUE #141: Use Handlebars {{each}} etc. to replace need for long HTML strings in JS
http://handlebarsjs.com/builtin_helpers.html

Joe-Brady:

Fixes #141 
Fixes #140 (barring very small changes required to auto-display ticks once the builds are approved)
Fixup for Localize variables
Adds image to be displayed when site is shared on social media
Adds some scroll arrows on Latest page for mobile devices/when we have too many platforms to display on the horizontal selector. (Was already scrollable, but this makes it completely obvious to new users that you can scroll horizontally)

##### Checklist

- [x] `npm test` passes


src/handlebars/index.handlebars
```diff
@@ -17,9 +17,11 @@
         <i id="dl-icon" class="fa fa-download" aria-hidden="true"></i>
         <span>Latest build</span>
         <i id="dl-icon-2" class="fa fa-arrow-circle-o-right hide" aria-hidden="true"></i>
-        <br>
       </div>
-      <div class="dl-description"><span id="dl-version-text" class="small-dl-text"></span></div>
+      <div class="dl-description">
+        <var id="dl-version-text" class="small-dl-text" release-version></var>
+        <span id="jck-approved-tick" class='tick hide' data-tooltip="This build is JCK approved."><img src='./dist/assets/tick.png' alt="This build is JCK approved."></span>
+      </div>
```
karianna: May need to provide a link to JCK

src/handlebars/releases.handlebars
```diff
@@ -34,12 +34,78 @@
       </table>
 
       <div id="latest-downloads-container">
-        <h2 style="margin-bottom: 0.5rem;">Select a platform</h2>
-        <h3 id="latest-scroll-text" class="hide">(Scroll horizontally)</h3>
+        <h2 id="latest-select-text" style="margin-bottom: 0.5rem;">Select a platform</h2>
+        <h3 id="latest-scroll-text" class="hide">Scroll to select a platform</h3>
+        <h1 id="latest-scroll-arrows" class="animated pulse hide">
+          <i id="latest-selector-left-scroll" class="fa fa-arrow-circle-left" aria-hidden="true"></i>
```
karianna: prefer em over i for tags?

Joe-Brady: https://developer.mozilla.org/en/docs/Web/HTML/Element/em

Just had a read of this as I wasn't aware they were essentially the same thing...
In this case, they are not being used to make the contents italic or emphasised - purely to contain the FontAwesome arrows and other icons.

The FontAwesome site recommends using <i> tags, so I had just followed that, e.g.:
 http://fontawesome.io/icon/arrow-circle-o-right/

src/handlebars/index.handlebars
Joe-Brady: Ok, I'll bear that in mind - would be useful, agreed, although will need to have a think about the best way to do that as the tick icons are already nested inside an `<a>` tag.

So perhaps javascript to say 'when the user clicks a shield, go to a new tab with the JCK link'.

What would be the best link to go to?

src/handlebars/releases.handlebars
karianna: Yeah I think you're OK with FontAwesome's advice in this case.

src/handlebars/index.handlebars
karianna: https://en.wikipedia.org/wiki/Technology_Compatibility_Kit

Issue #111, opened 2017-05-18
Add IE compatibility


Joe-Brady:
Work out what was causing the original issue with IE and fix

Joe-Brady:
Adding a line to set the HTTP request header as allowing cross-origin requests resolves this problem.

However, on resolving this issue, about 4 more issues sprang up related to IE compatibility. It seems that some fairly fundamental things just don't work in IE. As a result, important JS features that support things like the Latest page, the orange banner, and JSON formatting, get broken in IE.

I think that I could end up spending a lot of time trying to find workarounds for so many fundamental features might be fighting a losing battle, when I could be spending the time on more valuable contributions!

I'm closing this issue in favour of keeping the current 'doesn't work in IE, please use another browser' type message.

karianna:
Not so fast :-D.  I suspect that once we're up and running that we'll have plenty of IE users (even older than 11!) to our site.  We'll need to run Google Analytics to be sure of course.

What are the issues that are occurring?  A common technique is to detect IE and special case some JS or CSS for it.  There are a few libs which really help with this stuff as well.

Joe-Brady:
I'll explore this in depth when there's next time - thanks for highlighting that there are libraries to help with this stuff!

allandequeiroz:
I've been talking with some front-end guys and they gave me some hints on how to solve this. When I finish that DISPLAY stuff on tests I'll give it a try.

Joe-Brady:
@allandequeiroz Brilliant - thank you

PR #146, opened 2017-06-17
IE support

allandequeiroz:
Hello guys, to support IE I used two techniques here.

- Transpile : Since we're using new language features like **const** transpiling was necessary to generate code that IE manage to work with.

- Polyfill : Here, instead of using a full library that includes lots of unnecessary things that we don't need slowing down the website, I used a more precise way to solve it, including just what we need, eg:
[findIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex)

So far we're ok with IE [10,11]. Do we need to support lower versions? Looks like that for older ones even look and feel breaks, if so this will ned a much bigger work with CSS.

closes: #111


gdams:
staging: https://staging.adoptopenjdk.net

allandequeiroz:
Cool!

Joe-Brady:
Hi @allandequeiroz, huge thank you for the work!

I haven't checked on IE yet, but the changes seem to have broken the site on other browsers now (I've tested Chrome and Firefox on Mac so far...)

Something to do with `document.write` and Cloudflare... Any immediate thoughts on what this could be? I'm going to guess it's something to do with the changes to `function loadReleasesJSON(){}`, but really not sure!

screen shot 2017-06-18 at 12 35 58

allandequeiroz:
Hey @Joe-Brady, no worries, I saw the same warnings here, then I've checked live and same thing. 

And since everything was apparently working fine, I decided just to ignore them.

allandequeiroz:
What about range, IE [10,11] are enough? 

That was another assumption of mine. The target audience is composed of tech related people then maybe even that they are using IE let's say by company enforcement or some, they at least won't be using super old versions.

Do we have any analytics data about that?

Joe-Brady:
@allandequeiroz I just had a look at our Analytics page... in the all-time data so far, there have been very few IE visitors, and all of them have been on IE 11.

I think you're right that the target audience are very unlikely to be using a browser older than IE 10!

PR #148, opened 2017-06-18
Sort Latest and Archive platforms into specified order, add JCK tick links - Fixes #97, Fixes #147

### LINKED ISSUE #97: Make Archive and Latest platforms match the order of platforms array
Currently, the platform arrays are built and displayed alphabetically.

Find a way to easily specify an order (probably using a new property in the global platforms array objects), and have the website display the platforms in this order.

### LINKED ISSUE #147: Add JCK link on shield icon click
karianna:
May need to provide a link to JCK

Joe-Brady:
Ok, I'll bear that in mind - would be useful, agreed, although will need to have a think about the best way to do that as the tick icons are already nested inside an <a> tag.

So perhaps javascript to say 'when the user clicks a shield, go to a new tab with the JCK link'.
Will need to prevent bubbling as well otherwise the user would also start the file downloading.

What would be the best link to go to? 

karianna:
https://en.wikipedia.org/wiki/Technology_Compatibility_Kit

Joe-Brady:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/0-global.js
+++ b/src/js/0-global.js
@@ -98,6 +98,26 @@ function getOfficialName(searchableName) {
   return (lookup[searchableName].officialName);
 }
 
+function getPlatformOrder(searchableName) {
+  var index = platforms.findIndex(function(x) {
+    return x.searchableName == searchableName;
+  });
+  return index;
+}
+
+function orderPlatforms(inputArray) {
+  function compareOrder(a,b) {
+    if (a.thisPlatformOrder < b.thisPlatformOrder)
+      return -1;
+    if (a.thisPlatformOrder > b.thisPlatformOrder)
+      return 1;
+    return 0;
+  }
+
+  var orderedArray = inputArray.sort(compareOrder);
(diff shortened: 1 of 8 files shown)


src/js/0-global.js
```diff
@@ -98,6 +98,26 @@ function getOfficialName(searchableName) {
   return (lookup[searchableName].officialName);
 }
 
+function getPlatformOrder(searchableName) {
+  var index = platforms.findIndex(function(x) {
```
karianna: what is an 'x' :-)

src/js/0-global.js
```diff
@@ -98,6 +98,26 @@ function getOfficialName(searchableName) {
...
+    return x.searchableName == searchableName;
+  });
+  return index;
+}
+
+function orderPlatforms(inputArray) {
+  function compareOrder(a,b) {
```
karianna: what is an a and a b :-)

PR #153, opened 2017-06-22
Move platforms array to .json file

Joe-Brady:
This change prepares for an API code change, so that the platforms array can be shared between the website and API.

##### Checklist

- [x] `npm test` passes


gdams:
is it worth renaming `platforms.json` to `config.json` that way it can be used as a generic JSON file for all configurations? not sure if there is anything else that would be useful to store in JSON format

Joe-Brady:
Sure, I'll do that

PR #164, opened 2017-07-04
Add variant switcher (hidden, not activated)

Joe-Brady:
##### Checklist

- [x] `npm test` passes

[STAGING](https://staging.adoptopenjdk.net/) - no visual differences for now, though!

In the future we will need to be able to provide, for example, OpenJDK 9 as well as OpenJDK 8.

This PR contains the code that will in the future allow users to switch between these 'variants' on any of the main pages, and remain on that 'variant' while navigating around the site.

- A drop-down list provides the variant options.
- This drop-down list is populated by the new 'variants' array within `config.json`.
- The website defaults to the first entry in this 'variants' array - currently OpenJDK 8.
- When a new variant is selected, the current webpage reloads with the new variant's JSON file.
- The URL includes a `?variant=openjdk8` 'search' or 'query', so copying the URL to share the current page will link to the correct variant. If this is not included in a link you share, the website will default to the first entry in the array as before.

---

**TO ACTIVATE AT A LATER DATE:** (after @gdams has time to update everything, including the repos, to use the variant name, e.g. openjdk8-releases / openjdk8-nightly)
- Complete the TODO comments just before the `loadJSON` function in `archive.js`, `index.js`, `releases.js`, and `nightly.js`.

---

**HOW TO ADD OTHER VARIANTS WHEN AVAILABLE, AND TO DISPLAY THE SWITCHER:**
- The code assumes that two new repositories will exist for each variant within the AdoptOpenJDK org on GitHub, containing the usual four JSON files:
`latest_release.json`
`releases.json`
`latest_nightly.json`
`nightly.json`
- These repositories should be called, for example, `openjdk9-releases` and `openjdk9-nightly`.
- Add a new entry to the 'variants' array in `config.json`.
- In `styles-1-large-main.scss`, `select#variant-selector` currently has the property `display: none;`. This should be changed to `display: block;` to display the switcher.

---

# SCREENSHOTS

screen shot 2017-07-02 at 12 36 31
screen shot 2017-07-02 at 12 37 13
screen shot 2017-07-02 at 12 36 02
screen shot 2017-07-02 at 12 36 57


src/js/0-global.js
```diff
@@ -179,3 +184,62 @@ function setTickLink() {
...
+  }
+}
+
+function getQueryByName(name) {
+  var url = window.location.href;
+  name = name.replace(/[\[\]]/g, '\\$&');
+  var regex = new RegExp('[?&]' + name + '(=([^&#]*)|&|#|$)');
+  var results = regex.exec(url);
+  if (!results) return null;
+  if (!results[2]) return '';
```
karianna: 2 is a magic number

PR #166, opened 2017-07-07
WIP: Add dynamic installation page

Joe-Brady:
[ON STAGING - Installation](https://staging.adoptopenjdk.net/installation.html)
##### Checklist

- [x] `npm test` passes

- The installation page is now dynamically populated, similar to the other main pages.
- The user has a drop-down list of platforms to choose from.
- If the user's OS is recognised and matches a supported platform, the relevant platform is selected and displayed on page load - UNLESS a platform has already been specified (via the URL, if for example you send someone a link to that specific platform installation instructions).
- The installation instructions now contain the EXACT commands that users will need to run, to install the latest binary for their selected platform (thanks for the suggestion @gdams)
- `config.js` contains the 'template' unzip, path, and checksum commands for each platform.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -12,9 +12,6 @@ globals:
   detectOS: true
   loadJSON: true
   errorContainer: true
-  onIndexLoad: true
-  onNightlyLoad: true
-  onReleasesLoad: true
   getSearchableName: true
   getOfficialName: true
   getBinaryExt: true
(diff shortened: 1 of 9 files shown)


Joe-Brady:
Travis build seems to have not started, but it passes eslint :)

Joe-Brady:
Added copy buttons for codeblocks

Joe-Brady:
Seems to be a problem with the code highlighting - it doesn't work if you hard-refresh, but it does if you do a standard refresh. Will look into this, please don't merge yet.

Joe-Brady:
Seems to be a problem with the highlighting - can't seem to get `hljs.initHighlightingOnLoad();` to load at the right time, even with all the usual techniques.

PR #172, opened 2017-07-26
Add Installers to Home Page and Archive.

### LINKED ISSUE #170: Add Installers to home page and Archive
After the first installers (Win and Mac only at the moment) become available, they should automatically appear on the Latest page alongside the binaries, perhaps with a little tweaking required when that happens.

Once this looks 100%, the same functionality needs to be extended to the home page (perhaps just having the installer replace the binary on the big blue button when available), and to the Archive page.

EnriqueL8:
Fixes #170 
[ON STAGING](https://staging.adoptopenjdk.net) 

##### Checklist

- [x] `npm test` passes

Added capability for installers to be displayed on the Archive page and set capability and preference for installer on the main home page button.

### DIFF
--- a/src/handlebars/archive.handlebars
+++ b/src/handlebars/archive.handlebars
@@ -23,11 +23,9 @@
 
     <div id="archive-list" class="hide">
       <select id="variant-selector"></select>
-
       <div id="pagination-container"></div>
       <table class='archive-container'>
         <tbody id='archive-table-body'>
-
           <!-- BEGIN HANDLEBARS LIVE-RENDERED TEMPLATE -->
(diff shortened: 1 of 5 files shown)


karianna:
Can we see those hosted on staging?

EnriqueL8:
@karianna  They are hosted on staging.

karianna:
Got it, I assume installers isn't showing because we don't have any?

EnriqueL8:
Yes, when we have the installers it will show a column with the current installers as such : ( ignore the .tar.gz extension it is just for testing)
image

gdams:
Great first contribution here!

PR #174, opened 2017-08-22
Configured links to enforce openjdk8  repo

EnriqueL8:
[ON STAGING](https://staging.adoptopenjdk.net) 
Setting archive and nightly build links to point to openjdk8 repositories. 

##### Checklist

- [x] `npm test` passes
- [X] contribution guidelines followed [here](https://github.com/AdoptOpenJDK/openjdk-website/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/handlebars/partials/menu.handlebars
+++ b/src/handlebars/partials/menu.handlebars
@@ -9,6 +9,7 @@
       <li><a href="./archive.html">Build archive</a></li>
       <li><a href="./nightly.html">Nightly builds</a></li>
       <li><a href="./installation.html">Installation</a></li>
+      <li><a href="http://blog.adoptopenjdk.net/">Blog</a></li>
       <li><a target="_blank" href="http://api.adoptopenjdk.net">API</a></li>
       <li class="submenu"><a>Further information</a>
         <ul>
--- a/src/js/archive.js
+++ b/src/js/archive.js
@@ -72,11 +72,24 @@ function buildArchiveHTML(releasesJson) {
       // firstly, check if the platform name is recognised...
       if(ASSETOBJECT.thisPlatform) {
 
+        // if the filename contains both the platform name and the matching INSTALLER extension, add the relevant info to the asset object
+        ASSETOBJECT.thisInstallerExtension = getInstallerExt(ASSETOBJECT.thisPlatform);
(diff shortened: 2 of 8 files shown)


src/handlebars/partials/menu.handlebars
```diff
@@ -9,6 +9,7 @@
       <li><a href="./archive.html">Build archive</a></li>
       <li><a href="./nightly.html">Nightly builds</a></li>
       <li><a href="./installation.html">Installation</a></li>
+      <li><a href="http://blog.adoptopenjdk.net/">Blog</a></li>
```
karianna: https?

src/js/archive.js
```diff
@@ -18,7 +18,7 @@ function populateArchive() {
 
     // var repoName = (variant + '-releases');
 
-    loadJSON('openjdk-releases', 'releases', function(response) {
+    loadJSON('openjdk8-releases', 'releases', function(response) {
```
karianna: OK we'll have to make this work for OpenJDK 9 as well at some point, but that's a later fix.

src/js/index.js
```diff
@@ -52,55 +52,66 @@ function buildHomepageHTML(releasesJson) {
...
+      var uppercaseOSname = null;
       // firstly, check if a valid searchableName has been returned (i.e. the platform is recognised)...
       if(thisPlatform) {
 
         // secondly, check if the file has the expected file extension for that platform...
         // (this filters out all non-binary attachments, e.g. SHA checksums - these contain the platform name, but are not binaries)
-        var thisBinaryExtension = getBinaryExt(thisPlatform); // get the file extension associated with this platform
-        if(uppercaseFilename.indexOf(thisBinaryExtension.toUpperCase()) >= 0) {
-          var uppercaseOSname = OS.searchableName.toUpperCase();
+        var thisBinaryExtension = getBinaryExt(thisPlatform); // get the binary extension associated with this platform
```
karianna: Comments are probably not needed, function call is well named

src/js/archive.js
Joe-Brady: @karianna I have worked with Enrique just now to change this - I had set it up before I left to work with all variants that are defined inside `config.json`. So Enrique has now changed these lines of code to activate this system.

When OpenJDK 9 comes along, it's just a matter of choosing a naming convention for the new repositories, then adding a few lines to `config.json` to match this naming convention.

PR #220, opened 2018-03-14
add dynamic JCK status

gdams:
This PR allows us to dynamically set which platforms are JCK approved for a specific platform release.

## How will we mark binaries as JCK compliant?

We will have a jck.json file in a release repo. It will look something like this:
```JSON
{
  "jdk8u152-b16": {
    "PPC64LE_LINUX": "true"
  }
}
```

I have added a file here for testing: https://github.com/AdoptOpenJDK/openjdk8-openj9-releases/blob/master/jck.json.

NOTE that the CSS doesn't look fantastic on this. Perhaps @EnriqueL8 could take a look and pick this up later on but for now this is the important work.

## Staging
https://staging.adoptopenjdk.net/releases.html?variant=openjdk8-openj9#ppc64le_linux

### DIFF
--- a/src/js/0-global.js
+++ b/src/js/0-global.js
@@ -141,21 +141,24 @@ function detectOS() {
 // when using this function, pass in the name of the repo (options: releases, nightly)
 function loadJSON(repo, filename, callback) {
   var url = ('https://raw.githubusercontent.com/AdoptOpenJDK/' + repo + '/master/' + filename + '.json'); // the URL of the JSON built in the website back-end
-
   if(repo === 'adoptopenjdk.net') {
     url = (filename);
   }
-
   var xobj = new XMLHttpRequest();
-  xobj.open('GET', url, true);
+  xobj.open('GET', url, true)
   xobj.onreadystatechange = function () {
     if (xobj.readyState == 4 && xobj.status == '200') { // if the status is 'ok', run the callback function that has been passed in.
       callback(xobj.responseText);
     } else if (
       xobj.status != '200' && // if the status is NOT 'ok', remove the loading dots, and display an error:
       xobj.status != '0') { // for IE a cross domain request has status 0, we're going to execute this block fist, than the above as well.
-      loading.innerHTML = '';
-      document.getElementById('error-container').innerHTML = '<p>Error... there\'s a problem fetching the releases. Please see the <a href=\'https://github.com/AdoptOpenJDK/openjdk-releases/releases\' target=\'blank\'>releases list on GitHub</a>.</p>';
+        if (filename !== 'jck') {
+          document.getElementById('error-container').innerHTML = '<p>Error... there\'s a problem fetching the releases. Please see the <a href=\'https://github.com/AdoptOpenJDK/openjdk-releases/releases\' target=\'blank\'>releases list on GitHub</a>.</p>';
+          loading.innerHTML = '';
+        } else {
+          loading.innerHTML = '';
+          callback(null)
+        }
     }
   };
   xobj.send(null);
@@ -260,9 +263,9 @@ function setVariantSelector() {
     if(!variant) {
       variant = variants[0].searchableName;
     }
-    
+
     variantSelector.value = variant;
-    
+
     if(variantSelector.value === '') {
       var op = new Option();
       op.value = 'unknown';
--- a/src/js/0-global.js
+++ b/src/js/0-global.js
@@ -145,7 +145,7 @@ function loadJSON(repo, filename, callback) {
     url = (filename);
   }
   var xobj = new XMLHttpRequest();
-  xobj.open('GET', url, true)
+  xobj.open('GET', url, true);
   xobj.onreadystatechange = function () {
     if (xobj.readyState == 4 && xobj.status == '200') { // if the status is 'ok', run the callback function that has been passed in.
       callback(xobj.responseText);
--- a/src/js/archive.js
+++ b/src/js/archive.js
@@ -107,7 +107,7 @@ function buildArchiveHTML(releasesJson, jckJSON) {
           if (Object.keys(jckJSON).length == 0) {
             ASSETOBJECT.thisVerified = false;
           } else {
-            if (jckJSON[eachRelease.name].hasOwnProperty(ASSETOBJECT.thisPlatform) ) {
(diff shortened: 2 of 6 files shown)


sxa555:
@lumpfish @karianna Is "JCK compliant" the correct term here? Should it be "JCK Certified"? "JCK Validated"? Otherwise looks good for now.

DanHeidinga:
I'd vote for `JCK Certified`.  The JCK certifies the implementation is compatible with the standard.

gdams:
The production jenkins job has been updated to remove the banner for the next website build so this should be good to land

gdams:
okay so there appears to be one last bug on https://staging.adoptopenjdk.net/archive.html?variant=openjdk8-openj9. I will fix this and push in the morning.

karianna:
I'm curious about the need for an explicit file here.  Can we not give the mark of approval based on the successful execution of a Pipeline job?

gdams:
So the idea is that we will automate adding binaries to the `jck.json` file when the JCK jobs are complete and passing but for now we can just manually add the binaries :)

karianna:
OK so we're putting in a manual hook for what will be pipelined....  OK, let me review the code now :-).

src/js/0-global.js
```diff
@@ -141,21 +141,24 @@ function detectOS() {
...
 function loadJSON(repo, filename, callback) {
   var url = ('https://raw.githubusercontent.com/AdoptOpenJDK/' + repo + '/master/' + filename + '.json'); // the URL of the JSON built in the website back-end
-
   if(repo === 'adoptopenjdk.net') {
     url = (filename);
   }
-
   var xobj = new XMLHttpRequest();
-  xobj.open('GET', url, true);
+  xobj.open('GET', url, true)
```
karianna: Can I have my `;` back please :-)

gdams:
yeah I will create a jenkins job that is able to edit that JSON file at some point but I think that manually setting JCK compliance is good for now

src/js/0-global.js
```diff
@@ -141,21 +141,24 @@ function detectOS() {
...
     } else if (
       xobj.status != '200' && // if the status is NOT 'ok', remove the loading dots, and display an error:
       xobj.status != '0') { // for IE a cross domain request has status 0, we're going to execute this block fist, than the above as well.
-      loading.innerHTML = '';
-      document.getElementById('error-container').innerHTML = '<p>Error... there\'s a problem fetching the releases. Please see the <a href=\'https://github.com/AdoptOpenJDK/openjdk-releases/releases\' target=\'blank\'>releases list on GitHub</a>.</p>';
+        if (filename !== 'jck') {
+          document.getElementById('error-container').innerHTML = '<p>Error... there\'s a problem fetching the releases. Please see the <a href=\'https://github.com/AdoptOpenJDK/openjdk-releases/releases\' target=\'blank\'>releases list on GitHub</a>.</p>';
+          loading.innerHTML = '';
+        } else {
+          loading.innerHTML = '';
```
karianna: So if we error out trying to get a JCK release we say nothing?

src/js/archive.js
```diff
@@ -29,7 +29,13 @@ function populateArchive() {
 
       // if there are releases prior to the 'latest' one (i.e. archived releases)...
       if (typeof releasesJson[0] !== 'undefined') {
-        buildArchiveHTML(releasesJson);
+        loadJSON(repoName, 'jck', function(response_jck) {
+          var jckJSON = {}
+          if (response_jck !== null){
+            jckJSON = JSON.parse(response_jck)
+          }
+          buildArchiveHTML(releasesJson, jckJSON);
```
karianna: Couldn't we make JCK status part of the releasesJson?  That is, merge it?

src/js/archive.js
```diff
@@ -123,7 +137,15 @@ function buildArchiveHTML(releasesJson) {
             ASSETOBJECT.thisBinarySize = Math.floor((eachAsset.size)/1024/1024);
             ASSETOBJECT.thisChecksumLink = (eachAsset.browser_download_url).replace(ASSETOBJECT.thisBinaryExtension, '.sha256.txt');
             ASSETOBJECT.thisPlatformOrder = getPlatformOrder(ASSETOBJECT.thisPlatform);
-            ASSETOBJECT.thisVerified = false;
+            if (Object.keys(jckJSON).length == 0) {
```
karianna: DRY block above

src/js/releases.js
```diff
@@ -19,7 +19,13 @@ function populateLatest() {
     loadJSON(repoName, 'latest_release', function(response) {
       var releasesJson = JSON.parse(response);
       if (typeof releasesJson !== 'undefined') { // if there are releases...
-        buildLatestHTML(releasesJson);
+        loadJSON(repoName, 'jck', function(response_jck) {
```
karianna: DRY block with archive.js

src/js/releases.js
```diff
@@ -66,7 +72,15 @@ function buildLatestHTML(releasesJson) {
       ASSETOBJECT.thisLogo = getLogo(ASSETOBJECT.thisPlatform);
       ASSETOBJECT.thisPlatformOrder = getPlatformOrder(ASSETOBJECT.thisPlatform);
       ASSETOBJECT.thisOfficialName = getOfficialName(ASSETOBJECT.thisPlatform);
-      ASSETOBJECT.thisVerified = false;
+      if (Object.keys(jckJSON).length == 0) {
```
karianna: DRY block with archive.js

src/scss/styles-4-releases.scss
```diff
@@ -142,6 +142,11 @@ $tableblue-even: #1F2F4E;
   color: rgba(255, 255, 255, .7);
 }
 
+.tick img {
```
karianna: I don't see this tick on staging?  Did you forget to commit the image file?  Probably better just to use a font for the tick?

src/js/0-global.js
gdams: the thought behind this is that not all repos will neccessarily have a jck.json file

src/js/archive.js
gdams: Not very easily. It would make our website-backend job a lot more complicated

src/scss/styles-4-releases.scss
gdams: It's not there atm because our latest release just got deleted and is being re built. It worked before.

PR #223, opened 2018-03-16
Update support page quality statements around JCK testing 

gdams:
##### Checklist

- [x] `npm test` passes

Thanks @mstoodle for the wording!

Staging: https://staging.adoptopenjdk.net/support.html


sxa555:
It's tempting, while the subset is so small, to explicitly enumerate on that page which ones we currently have passed the compliance tests on, since I'm not sure I'd easily be able to find a compliant build without a bit of searching as a user just coming to the web site, but if that's going to be added it should be a separate PR I suspect :-)

PR #235, opened 2018-04-03
a few bug fixes + tweaks

### LINKED ISSUE #228: Unfriendly error message when selecting the "Amber" variant
Raised by @tellison If you go to https://adoptopenjdk.net/?variant=amber you get the message:
```
Error... there's a problem fetching the releases. Please see the releases list on GitHub.
```
which is rather unfriendly. We should replace this with `there are no releases` somehow to make it clear that there isn't a fault in the website.

gdams:
##### Checklist

- [x] `npm test` passes

closes: https://github.com/AdoptOpenJDK/openjdk-website/issues/228

This PR fixes #228 but also tidies up the way that the JCK tick appears (making it more inline)

staging: https://staging.adoptopenjdk.net

### DIFF
--- a/src/js/0-global.js
+++ b/src/js/0-global.js
@@ -153,7 +153,15 @@ function loadJSON(repo, filename, callback) {
       xobj.status != '200' && // if the status is NOT 'ok', remove the loading dots, and display an error:
       xobj.status != '0') { // for IE a cross domain request has status 0, we're going to execute this block fist, than the above as well.
         if (filename !== 'jck') {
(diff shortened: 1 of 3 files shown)


src/js/0-global.js
```diff
@@ -153,7 +153,15 @@ function loadJSON(repo, filename, callback) {
       xobj.status != '200' && // if the status is NOT 'ok', remove the loading dots, and display an error:
       xobj.status != '0') { // for IE a cross domain request has status 0, we're going to execute this block fist, than the above as well.
         if (filename !== 'jck') {
-          document.getElementById('error-container').innerHTML = '<p>Error... there\'s a problem fetching the releases. Please see the <a href=\'https://github.com/AdoptOpenJDK/openjdk-releases/releases\' target=\'blank\'>releases list on GitHub</a>.</p>';
+          console.log(xobj.status)
```
karianna: Do we need to log this?  If so then a more descriptive log line would be better.

PR #262, opened 2018-06-20
Update searchableName for x64_linux large heap build

smlambert:
Remove the _ to match the value passed to website-backend Release.sh script.  This corrects the fact the the x64_linuxLH builds are being posted to the website under the heading x64_linux builds (though the artifact is named properly).


smlambert:
screen shot 2018-06-20 at 7 40 33 am

Correctly differentiates between Linux x64 and Linux x64 large heap builds on staging server now.

PR #266, opened 2018-07-11
Read new repositories

johnoliver:
Modify code to first check the new `openjdk<x>-binaries` repos to look for releases, if none can be found in those locations default to the old repos.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/0-global.js
+++ b/src/js/0-global.js
@@ -9,7 +9,7 @@ var variantSelector = document.getElementById('variant-selector');
 var platformSelector = document.getElementById('platform-selector');
 
 if(jvmVariant === undefined || jvmVariant === null) {
-  jvmVariant = "hotspot";
+  jvmVariant = 'hotspot';
 }
 
 function setLookup() {
@@ -144,7 +144,7 @@ function detectOS() {
 }
 
 function toJson(response) {
-  while (typeof response === "string") {
+  while (typeof response === 'string') {
     try {
       response = JSON.parse(response)
     } catch (e) {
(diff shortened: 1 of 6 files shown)


src/js/0-global.js
```diff
@@ -138,6 +143,47 @@ function detectOS() {
...
+  }
+  return response
+}
+
+// load latest_nightly.json/nightly.json/releases.json/latest_release.json files
+// This will first try to load from openjdk<X>-binaries repos and if that fails
+// try openjdk<X>-release, i.e will try the following:
+
+// https://github.com/AdoptOpenJDK/openjdk10-binaries/blob/master/latest_release.json
+// https://github.com/AdoptOpenJDK/openjdk10-releases/blob/master/latest_release.json
```
karianna: This is only for 'release' level builds or 'nightly' level builds?

johnoliver: No, those are just examples, in fact at the moment I have only pushed nightlys, so only nightlys will be found, but either should work

src/js/0-global.js
```diff
@@ -222,13 +255,21 @@ function setTickLink() {
...
 }
 
-function setUrlQuery(name, newValue) {
-  if(window.location.search.indexOf(name) >= 0) {
-    var currentValue = getQueryByName(name);
-    window.location.search = window.location.search.replace(currentValue, newValue);
-  }
-  else {
-    window.location.search += (name + '=' + newValue);
+function setUrlQuery() {
```
karianna: Might be nice to show a sample in a comment here

src/js/0-global.js
```diff
@@ -265,6 +306,8 @@ function persistUrlQuery() {
   });
 }
 
+const versionMatcher=/(openjdk\d+)-([a-zA-Z0-9]+)/;
```
karianna: Declare at the top if it's global?

src/js/archive.js
```diff
@@ -10,47 +10,36 @@ function onArchiveLoad() {
...
-          buildArchiveHTML(releasesJson, jckJSON);
-        });
-      } else { // if there are no releases (beyond the latest one)...
-        // report an error, remove the loading dots
-        loading.innerHTML = '';
-        errorContainer.innerHTML = '<p>There are no archived releases yet! See the <a href=\'./releases.html?variant=' + variant + '\'>Latest release</a> page.</p>';
+      if (response !== null) {
+        buildArchiveHTML(response);
+      } else {
+        if(oldRepo) {
```
karianna: Not sure why this logic is wrapped in oldRepo?

johnoliver: The sequence is we check the new repo, if there are no files then check the old one, this then says if we find nothing from the old, display an error

PR #275, opened 2018-07-26
Updating the AdoptOpenJDK website to include the jvm variant.

adamfarley:
This makes the links consistent and improves the accuracy of error
messages.

See issue https://github.com/AdoptOpenJDK/openjdk-website/issues/276

Signed-off-by: Adam Farley <adam.farley@uk.ibm.com>

### DIFF
--- a/src/js/archive.js
+++ b/src/js/archive.js
@@ -29,7 +29,7 @@ function populateArchive() {
       // if there are no releases (beyond the latest one)...
       // report an error, remove the loading dots
       loading.innerHTML = '';
-      errorContainer.innerHTML = '<p>There are no archived releases yet! See the <a href=\'./releases.html?variant=' + variant + '\'>Latest release</a> page.</p>';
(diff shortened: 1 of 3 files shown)


adamfarley:
Looks good on the staging server. Problem is resolved and the links are not consistent. If the jdk9 on openj9 nightly page says to check the latest releases page, it will link to the openj9 latest releases page rather than defaulting to the hotspot one.

@sxa555 - Can you review and merge when happy?

sxa555:
Yep I'm happy with that after having a play on the staging server - thanks for the fix

SueChaplain:
Hmm .. .the message doesn't work too well for the old OpenJ9 variant pages: https://adoptopenjdk.net/releases.html?variant=openjdk8-openj9

johnoliver:
How did you get that link? It should be https://adoptopenjdk.net/releases.html?variant=openjdk8&jvmVariant=openj9 and when I look through pages for openjdk8+j9  I get actual results and don't see an error message.

SueChaplain:
@johnoliver The OpenJ9 website was pointing to that link (https://www.eclipse.org/openj9/) .. we're just not up to date.  Don't worry, we're fixing now. Doesn't seem any point in trying to patch the message on the old page.

PR #303, opened 2018-09-28
change java 11 banner to use proper banner

gdams:
##### Checklist

- [x] `npm test` passes

This might not render well on staging as it already uses a different banner but this will work in production.


gdams:
correct! :)

Issue #283, opened 2018-08-14
Modify platform detection so that the openj9 large heap is not the default on Linux/x64


sxa:
1. Go to adoptopenjdk.net from a Linux/x64 browser
2. Select openjdk8-openj9 from the dropdown
3. Large heap build is selected as the default - it should be the standard compressed refs build

gdams:
closing as PR landed

sxa555:
@gdams The default when you go to it on Linux is currently the LH build - it points to https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u202-b08_openj9-0.12.1/OpenJDK8U-jdk_x64_linux_openj9_linuxXL_8u202b08-openj9-0.12.1.tar.gz - this is not fixed

Hello71:
Seems to work fine now.

PR #322, opened 2018-10-05
WIP Re-add LinuxLH jdk8 builds & stop them being default

sxa555:
(Hopefully) Fixes https://github.com/AdoptOpenJDK/openjdk-build/issues/610 and https://github.com/AdoptOpenJDK/openjdk-website/issues/283

### DIFF
--- a/src/json/config.json
+++ b/src/json/config.json
@@ -69,7 +69,19 @@
       "installCommand": "tar -xf FILENAME",
       "pathCommand": "export PATH=$PWD/DIRNAME/bin:$PATH",
       "checksumCommand": "sha256sum FILENAME",
-      "osDetectionString": "Linux Mint Debian Fedora FreeBSD Gentoo Haiku Kubuntu OpenBSD Red Hat RHEL SuSE Ubuntu Xubuntu hpwOS webOS Tizen"
+      "osDetectionString": "not-to-be-detected"
+    },
+    {
+      "officialName": "Linux x64 Large Heap",
+      "searchableName": "X64_LINUXLH",
+      "logo": "linux.png",
+      "binaryExtension": ".tar.gz",
+      "installerExtension": ".run",
+      "architecture": "64",
+      "installCommand": "tar -xf FILENAME",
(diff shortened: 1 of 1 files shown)


src/json/config.json
```diff
@@ -69,7 +69,19 @@
       "installCommand": "tar -xf FILENAME",
       "pathCommand": "export PATH=$PWD/DIRNAME/bin:$PATH",
       "checksumCommand": "sha256sum FILENAME",
-      "osDetectionString": "Linux Mint Debian Fedora FreeBSD Gentoo Haiku Kubuntu OpenBSD Red Hat RHEL SuSE Ubuntu Xubuntu hpwOS webOS Tizen"
+      "osDetectionString": "not-to-be-detected"
+    },
```
karianna: I'm not sure this is right (adding a new entry)?  We changed `X64_LINUXLH` to `LINUXXL` last week (so that's the Large Heap builds.

sxa555: @karianna The OpenJ9 JDK8 LH builds aren't showing up any more because they have the "old" name. Adding a separate section to cover both was my first stab at a fix to see if it worked (Hence the WIP prefix :-) ) - open to other options.

sxa555:
Based on the staging server this has resolved both issues and not impacted the visibility of the OpenJ9 Large Heap JDK11 buidls

src/json/config.json
karianna: ahhhhhhh.  Hmmm.

karianna: I'm worried that those who follow behind us are doomed to never understand why.  But you can't add comments in a json doc.....  OK - I'll let this through if we raise another issue to document this / fix upstream

sxa555: If we modify the scripts to produce the "new" artifact name for 8 (maybe that'll just happen) then we could remove it release once the next update comes out next month - would mean we'd lose the LH build from the main adoptopenjdk website for that one (to be historic) release but certainly simpler than trying to rename anything. I can keep the originating issue open until we resolve - will add a comment now.

sxa555: (Although I'm not sure why we renamed the artifact in the first place)

Issue #313, opened 2018-10-03
Add Javascript test(s)


karianna:
around the config.json file in particular

ParkerM:
When you mention config.json here are you hinting at something like a json schema definition that the config could be validated against in CI?

karianna:
Yes!

PR #330, opened 2018-10-07
Add config.json schema and validation task

ParkerM:
This relates to #313 

* Added a config.schema.json file and added some patterns to validate against. Some of the constraints are pretty loose but I didn't want to make too many assumptions about their formatting. I tried to base the schema descriptions on the contributing guide as much as possible.
* Created a `json-validate` task in the gulpfile that validates config.json against config.schema.json. It uses the assertion libraries to invoke test failures, and `ajv` (schema validator) and `plugin-error` were also added to devDependencies to support the validation task.

I wasn't sure how to add the task to the CI configuration without tacking it onto the existing npm test script, but it can be run locally if anyone wants to mess with various things to see the output. There's also a few defs I was iffy about so they're denoted by nearby `"$comment"` blocks in the schema.

Example sad path output after mucking with some config values (changing "64" to "65", removing required property, and changing .png to unsupported file type respectively):
```
[23:36:58] Using gulpfile ~\workspace\jdk\openjdk-website\gulpfile.js
[23:36:58] Starting 'json-validate'...
[23:36:58] Validating config.json against config.schema.json
[23:36:58] 'json-validate' errored after 46 ms
[23:36:58] AssertionError [ERR_ASSERTION] in plugin "json-validate"
Message:
    
data.platforms[5].architecture should be equal to one of the allowed values. Actual: "undefined",
data.platforms[6] should have required property 'checksumCommand'. Actual: "undefined",
data.platforms[7].logo should match pattern "^(.*\.(png|jpg|jpeg))$". Actual: "undefined"

Process finished with exit code 1
```
`Actual: "undefined"` was working as expected earlier (i.e. showing the problematic value), ~I believe it may be due to my pre-push switch from yarn to npm making a mess of my local env.~ Edit: nope, turns out it was accidental regression on my part, fixed in 9c7d2b6.
If anyone has questions or concerns about these changes please let me know!

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -165,7 +165,7 @@ gulp.task('json-validate', function () {
   const configJsonSchema = objFromJson('./src/json/config.schema.json');
   const configJson = objFromJson('./src/json/config.json');
 
-  const ajv = new Ajv({loadSchema: loadSchemaFn, allErrors: true, extendRefs: 'fail'});
+  const ajv = new Ajv({loadSchema: loadSchemaFn, allErrors: true, extendRefs: 'fail', verbose: true});
 
   return ajv.compileAsync(configJsonSchema)
     .then(validate =>
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,5 +1,9 @@
 const gulp = require('gulp');
 
+const fs = require('fs');
+const Ajv = require('ajv');
+const assert = require('assert');
+const PluginError = require('plugin-error');
 const runSequence = require('run-sequence');
 const browserSync = require('browser-sync').create();
 const concat = require('gulp-concat');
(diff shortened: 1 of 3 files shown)


karianna:
Great stuff!  You raise a good question as well.  What does `npm test` actually call.  OK, so digging in I think the _package.json_ determines what gets called when you call `npm test` or `npm start` etc.  So I think you have two choices in package.json:

1. make sure the `default` gulp task (which gets called by `npm`) start also includes your new `json-validate` gulp target as a dependency.
2. change the `npm test` to still call `npm run lint` but also to include a call to your specific gulp `json-validate`
3. Do both (preferred)!

karianna:
Also just noticed that you've got the commits from a previous PR that landed - probably need to rebase

ParkerM:
> Also just noticed that you've got the commits from a previous PR that landed - probably need to rebase

Whoops, good catch! That should be rebased now.

e9fc2d7 adds the validation task as a sequential prereq before the concurrent build/dist tasks in the gulp default (I assumed we'd want to fail fast if a config.json or schema rule is bad). Also added a `json-validate` npm script and appended it to `npm-test` (also sequentially to avoid log weirdness).

Issue #307, opened 2018-10-01
Advice on front page on choosing hotspot or openj9?


andrew-m-leonard:
For those who are just after an open source Java binary, they might not know the difference between hotspot and openj9, or whether it matters to them...or what it is! They may have just hit the https://adoptopenjdk.net/ page from google, should we give them some advice or pointers?
Should we just default to hotspot for them? or openj9? or display both at the same time (without a dropdown)? and a summary over choosing...?

karianna:
This really ties into providing a bit more of a wizard/persona approach.  I'd be happy to see a PR where Hotspot and OpenJ9 options sit side by side for each version / platform combo

ghenriks:
Talk of radio buttons, wizards, etc. are missing the point - most people are going to come to adoptopenjdk to simply download Java and don't care about the choices (in a large part because for most of them the differences won't matter).

If given the confusion of having to make a choice, and the worry of making the wrong choice, many of them will simple shrug their shoulders, decide Java isn't for them and instead go download Go, Rust, Dotnet Core, or any number of other language choices that don't require decision making to download and install.

So by all means continue to offer both builds (or even more if you want), but hide it so only those who really care / want it have to get exposted to it and make the front page a simply choose your release version and download.

Also, I would suggest going further and only offering the currently supported LTS versions on the initial page - inexperienced people shouldn't be offered versions that will only be supported for 6 months or less depending on download date.

ghenriks:
> Talk of radio buttons, wizards, etc. are missing the point 

Of course it now appears that all really doesn't matter given that you aren't even offering Windows users an installable, but instead expect them to somehow figure out how to change their path variable to point to wherever they forgot they unzipped the zip file to.

Given that I can never remember where Microsoft has hidden the path variable, I certainly am not going to be encouraged to work with a product where I have to look that up for each new version...

karianna:
@ghenriks This is a new community project that's been asked to deliver a very large amount of infrastructure and software to the ecosystem for $free and free as in use.  We are working on the issues that you mention (MSI installer is being shaken out - https://github.com/AdoptOpenJDK/openjdk-installer/pull/6) etc as well as rethinking the Ux of the website.  

We'd love to have your help!  Come and join us on the slack adoptopenjdk.net/slack.html

SeanC1083:
Help Me Choose link to popup info about both JVMs

[...]

Issue #331, opened 2018-10-08
Display link to JREs in download


ColinHebert:
Currently we do not display links to the JRE downloads.

The JREs are technically accessible through the API ( https://github.com/AdoptOpenJDK/openjdk11-binaries/blob/master/latest_release.json ), but not from the website.

karianna:
You can now get access to them in the archives.  We do need to think about making it default option alongside the JDK

sxa555:
They are also in the Installation page visible when you click the burger icon in the top top left of the site. I've also also proposing adding it to the main page as part of https://github.com/AdoptOpenJDK/openjdk-website/pull/339#issuecomment-434234799

SeanC1083:
@ddunn2 @karianna @sxa555 I am currently working on some redesigns that include ensuring that users have access to JRE downloads. I will likely share these designs later this week. (Repost from #339)

gdams:
closing as the website now serves both JDK's and JRE's

Issue #301, opened 2018-09-26
What are Hotspot and Eclipse OpenJ9?


janpio:
When visiting https://adoptopenjdk.net/ on Windows I get these download options:

![image](https://user-images.githubusercontent.com/183673/46078829-93472000-c195-11e8-930d-3a17ce2adfff.png)

What are Hotspot and Eclipse OpenJ9? Which should I choose and why?
(What is "amber"?)
I couldn't find any information on that page or "About".

EsRa13:
short google shows:

https://en.wikipedia.org/wiki/Comparison_of_Java_virtual_machines
https://www.eclipse.org/openj9/oj9_faq.html
...

janpio:
So the expectation should be for everyone to use Google to learn about this? I hope not.

karianna:
This is being resolved in https://github.com/AdoptOpenJDK/openjdk-website/pull/339

PR #339, opened 2018-10-17
Add radio buttons to website for variant selectors

ddunn2:
As per issue #307 I've added radio buttons in place of the drop down menu on the various pages it occurs. 

I've not implemented the `Help Me Choose` section yet as I was asked to just focus on the radio buttons for now, and there seems to be a separate issue for that section. 

I've tested these changes on: 

- Chrome
- Safari
- FireFox
- Internet Explorer
- Microsoft Edge
- iPhone
- iPad

### DIFF
--- a/src/handlebars/index.handlebars
+++ b/src/handlebars/index.handlebars
@@ -19,22 +19,21 @@
         <h3>1. Choose a Version</h3>
       </form>
       <form id="jvm-selector" class="btn-form">
-        <h3>2. Choose a JVM</h3>
+        <h3>2. Choose a JVM
+          <span class="help-jvm">Help Me Choose</span>
+          <div class="popup-content" id="jvm-popup">
+            <div class="popup-heading">
+              <h4>Choosing the right JVM</h4>
+            </div>
+            <div class="popup-text">
+              <p class="hotspot-text"><b>Hotspot</b> is the VM from the OpenJDK community. It is the most widely used VM today (e.g. It is used in Oracle's JDK) and is suitable for all workloads.</p>
+              <p>For more details see <a target="_blank" href="https://openjdk.java.net">OpenJDK</a>.</p>
+              <p class="openj9-text"><b>Eclipse OpenJ9</b> is the VM that was open sourced by IBM and is also suitable for running all workloads. This variant uses the same class libraries from OpenJDK, i.e. It is interchangeable from the Hotspot variant as far as Java programmers are concerned.</p>
+              <p>For more details see <a target="_blank" href="https://www.eclipse.org/openj9/">Eclipse OpenJ9</a>.</p>
+            </div>
+          </div>
+        </h3>
       </form>
-      <div class="popup" onclick="togglePopup()">
-        <span>Help Me Choose</span>
-      </div>
-      <div class="popup-content" id="jvm-popup">
-        <div class="popup-heading">
-          <h4>Choosing the right JVM</h4>
-        </div>
-        <div class="popup-text">
-          <p class="hotspot-text">Hotspot is the VM from the OpenJDK community. It is the most widely used VM today (e.g. It is used in Oracle's JDK) and is suitable for all workloads.</p>
-          <p>For more details see <a target="_blank" href="https://openjdk.java.net">OpenJDK</a>.</p>
-          <p class="openj9-text">Eclipse OpenJ9 is the VM that was open sourced by IBM and is also suitable for running all workloads. This variant uses the same class libraries from OpenJDK, i.e. It is interchangeable from the Hotspot variant as far as Java programmers are concerned.</p>
-          <p>For more details see <a target="_blank" href="https://www.eclipse.org/openj9/">Eclipse OpenJ9</a>.</p>
-        </div>
-      </div>
     </div>
 
     <div id='loading'><img src='dist/assets/loading_dots.gif' height='40' width='40'></div>
@@ -87,14 +86,6 @@
       <a target="_blank" href="https://calendar.adoptopenjdk.net"><h3 class="inline-block zero-margin">Meeting diary</h3></a>
     </div>
   </div>
-
-  <script type="text/javascript">
-  function togglePopup() {
-    var popup = document.getElementById('jvm-popup');
-    popup.classList.toggle('show');
-  }
-  </script>
-
 </main>
 
 {{> footer script='<script>onIndexLoad();</script>' script2='<script src="https://cdnjs.cloudflare.com/ajax/libs/platform/1.3.4/platform.min.js"></script>'}}
--- a/src/handlebars/index.handlebars
+++ b/src/handlebars/index.handlebars
@@ -23,16 +23,16 @@
       </form>
       <div class="popup" onclick="togglePopup()">
         <span>Help Me Choose</span>
-        <div class="popup-content" id="jvm-popup">
-          <div class="popup-heading">
-            <h4>Choosing the right JVM</h4>
-          </div>
-          <div class="popup-text">
-            <p>Hotspot is the VM from the OpenJDK community. It is the most widely used VM today (e.g. It is used in Oracle's JDK) and is suitable for all workloads.</p>
-            <p>For more details see <a target="_blank" href="https://openjdk.java.net">OpenJDK</a>.</p>
-            <p>Eclipse OpenJ9 is the VM that was open sourced by IBM and is also suitable for running all workloads. This variant uses the same class libraries from OpenJDK, i.e. It is interchangeable from the Hotspot variant as far as Java programmers are concerned.</p>
-            <p>For more details see <a target="_blank" href="https://www.eclipse.org/openj9/">Eclipse OpenJ9</a>.</p>
-          </div>
+      </div>
+      <div class="popup-content" id="jvm-popup">
+        <div class="popup-heading">
+          <h4>Choosing the right JVM</h4>
+        </div>
+        <div class="popup-text">
+          <p class="hotspot-text">Hotspot is the VM from the OpenJDK community. It is the most widely used VM today (e.g. It is used in Oracle's JDK) and is suitable for all workloads.</p>
+          <p>For more details see <a target="_blank" href="https://openjdk.java.net">OpenJDK</a>.</p>
+          <p class="openj9-text">Eclipse OpenJ9 is the VM that was open sourced by IBM and is also suitable for running all workloads. This variant uses the same class libraries from OpenJDK, i.e. It is interchangeable from the Hotspot variant as far as Java programmers are concerned.</p>
+          <p>For more details see <a target="_blank" href="https://www.eclipse.org/openj9/">Eclipse OpenJ9</a>.</p>
         </div>
       </div>
     </div>
@@ -87,6 +87,13 @@
       <a target="_blank" href="https://calendar.adoptopenjdk.net"><h3 class="inline-block zero-margin">Meeting diary</h3></a>
     </div>
   </div>
+
+  <script type="text/javascript">
+  function togglePopup() {
+    var popup = document.getElementById('jvm-popup');
+    popup.classList.toggle('show');
+  }
+  </script>
 
 </main>
 
--- a/src/handlebars/index.handlebars
+++ b/src/handlebars/index.handlebars
@@ -14,11 +14,11 @@
 
     <div id="error-container"></div>
 
-    <div class="btn-container">
-      <form id="jdk-selector" class="btn-form">
+    <div class="btn-container index-btn-container">
+      <form id="jdk-selector" class="jdk-btn-form btn-form">
         <h3>1. Choose a Version</h3>
       </form>
-      <form id="jvm-selector" class="btn-form">
+      <form id="jvm-selector" class="index-jvm-btn-form jvm-btn-form btn-form">
         <h3>2. Choose a JVM
           <span class="help-jvm">Help Me Choose</span>
           <div class="popup-content" id="jvm-popup">
--- a/src/handlebars/index.handlebars
+++ b/src/handlebars/index.handlebars
@@ -21,6 +21,20 @@
       <form id="jvm-selector" class="btn-form">
         <h3>2. Choose a JVM</h3>
       </form>
+      <div class="popup" onclick="togglePopup()">
+        <span>Help Me Choose</span>
+        <div class="popup-content" id="jvm-popup">
+          <div class="popup-heading">
+            <h4>Choosing the right JVM</h4>
+          </div>
+          <div class="popup-text">
+            <p>Hotspot is the VM from the OpenJDK community. It is the most widely used VM today (e.g. It is used in Oracle's JDK) and is suitable for all workloads.</p>
+            <p>For more details see <a target="_blank" href="https://openjdk.java.net">OpenJDK</a>.</p>
+            <p>Eclipse OpenJ9 is the VM that was open sourced by IBM and is also suitable for running all workloads. This variant uses the same class libraries from OpenJDK, i.e. It is interchangeable from the Hotspot variant as far as Java programmers are concerned.</p>
+            <p>For more details see <a target="_blank" href="https://www.eclipse.org/openj9/">Eclipse OpenJ9</a>.</p>
+          </div>
+        </div>
+      </div>
     </div>
 
     <div id='loading'><img src='dist/assets/loading_dots.gif' height='40' width='40'></div>
--- a/src/handlebars/index.handlebars
+++ b/src/handlebars/index.handlebars
@@ -28,7 +28,7 @@
             <div class="popup-text">
               <p class="hotspot-text"><b>Hotspot</b> is the VM from the OpenJDK community. It is the most widely used VM today (e.g. It is used in Oracle's JDK) and is suitable for all workloads.</p>
               <p>For more details see <a target="_blank" href="https://openjdk.java.net">OpenJDK</a>.</p>
-              <p class="openj9-text"><b>Eclipse OpenJ9</b> is the VM that was open sourced by IBM and is also suitable for running all workloads. This variant uses the same class libraries from OpenJDK, i.e. It is interchangeable from the Hotspot variant as far as Java programmers are concerned.</p>
+              <p class="openj9-text"><b>OpenJ9</b> is the VM from the Eclipse community.  It is an enterprise-grade VM designed for low memory usage and fast start-up and is used in IBM’s JDK.  It is also suitable for running all workloads.</p>
               <p>For more details see <a target="_blank" href="https://www.eclipse.org/openj9/">Eclipse OpenJ9</a>.</p>
             </div>
           </div>
--- a/src/handlebars/index.handlebars
+++ b/src/handlebars/index.handlebars
@@ -76,4 +76,4 @@
 
 </main>
 
-{{> footer script='<script src="https://cdnjs.cloudflare.com/ajax/libs/platform/1.3.4/platform.min.js"></script>' script2='<script>onIndexLoad();</script>' }}
+{{> footer script='<script>onIndexLoad();</script>' script2='<script src="https://cdnjs.cloudflare.com/ajax/libs/platform/1.3.4/platform.min.js"></script>'}}
--- a/src/handlebars/index.handlebars
+++ b/src/handlebars/index.handlebars
@@ -27,7 +27,7 @@
             </div>
             <div class="popup-text">
               <p class="hotspot-text"><b>Hotspot</b> is the VM from the OpenJDK community. It is the most widely used VM today (e.g. It is used in Oracle's JDK) and is suitable for all workloads.</p>
-              <p>For more details see <a target="_blank" href="https://openjdk.java.net">OpenJDK</a>.</p>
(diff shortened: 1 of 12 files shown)


karianna:
Adding some thoughts on the 9/10 choices.  We should label 8 and 11 as OpenJDK 8 (LTS) and OpenJDK 11 (LTS) to make it clear which are the LTS releases.  We could also add a "Which do I choose?" popover for the JDK choice with some useful info.  I'm happy to provide that wording in that helps?

SeanC1083:
@ddunn2 instead of numbering the option, could we made the download button inactive until a JVM selection is made. Another option would be to keep the download button active, but if users click the download button without making a JVM selection, we can have an alert or some text appear to let them know that a selection is needed. Once a JVM is selected it could pull up the proper JDK for users to download.

One example...

screen shot 2018-10-17 at 11 11 34 am

karianna:
3. I'm happy to look at an option for 8/9 OpenJDK moving to another page and seeing what the 'above the line' looks like that afterwards.

SeanC1083:
@ddunn2 I think the most desirable option would be to make the download button inactive until the user selects a JVM. We believe this will help the user understand that another selection is required to precede with getting the JDK download.

ddunn2:
@SeanC1083 The redlines were rather useful except they were in pixels from what I could see, and in the contributing guidelines for this project it states `Do not use pixels` so I attempted to create something similar using relative units, but comparing the two on different screens I definitely overshot the white spacing. 

@karianna The issue with moving the 8/9 options to a different page is the Javascript used to populate those buttons is used on every page. I could create a separate script that's just for the home page but still populate the other pages with all 4 options? 

@SeanC1083 I can look into making the button inactive however from what I understand about the code the button is generated only when you've made a choice. If there's no JVM selected a button would not be generated and we'd just have the 'loading dots' active until a selection is made. I could add some logic in to enable what you're suggesting if that's what we want to go for. @karianna  What's your opinion on this? I'm not sure how long this would take so I'd like to be certain this is the path we take :)

karianna:
@ddunn2 Having a separate script for the homepage sounds like a good idea.

You're correct in that the button is dynamically generated by the choices made.  So I don't think we can change it to be pre-populate but disabled.

SeanC1083:
@ddunn2 the 'loading dots' will work for now...given the upcoming deadline. 

So the radio options for JDK will default to OpenJDK 8, but the radio option for JVM will have no default, but will wait for the user to make a selection after which the download button will appear...?

karianna:
The JVM radio button should probably default to Hotspot.  I know we want to be as even as possible between the implementations but it is the default the majority will want.  And I'd prefer not to have to force extra clicks/steps in-between the user and their download.

SeanC1083:
@karianna sounds good. Thanks for your feedback.

ddunn2:
@karianna @SeanC1083 

I've pushed a couple more commits, the first deals with removing the obsolete JDK's from the home page. The second reduces the white space between the two radio sections. 

The script to remove the obsolete JDK options from the home page loops over and hides the obsolete options for the index page but continues to display them for the other pages. I found this to be the simplest solution. I did consider adding a separate script for generating the home page radio buttons but that would require a larger refactor. 

I also haven't addressed the 'above the line' item yet, I wasn't sure what it is you'd like users to be able to see without scrolling? 

I also left the numbers in on the `Choose a JDK` & `Choose a JVM`, is this what is wanted or would you prefer them to be removed?

karianna:
Let me force a staging site refresh and I'll get back to you :-)

karianna:
Latest screenshot.  It's definitely better!  I think we will need to add the "Which one Do I choose" style tooltips before we can release this.  The 'above the line' issue I think I can resolve in a separate PR.  A bit of whitespace tightening and making some of the description text also available via a tooltip may help there.

mac os 1425 by 1250

ddunn2:
image

Is this the tooltip you're referring to? If so I'll need the text you'd like to appear in the tooltip :)

Also I did play around with some white space tightening as well, there's definitely some areas we could get some space back, I just wasn't sure what you wanted above the line.

karianna:
@ddunn2 - yes that's one of the tooltips suggested text as follows:

_Hotspot is the VM from the OpenJDK community. It is the most widely used VM today (e.g. It is used in Oracle's JDK) and is suitable for all workloads.

For more details (https://openjdk.java.net)[OpenJDK].

Eclipse OpenJ9 is the VM that was open sourced by IBM. It has several features that can benefit your cloud/container workloads. NOTE: This variant uses the same class libraries from OpenJDK, i.e. It is interchangeable from the Hotspot variant as far as Java programmers are concerned._

From more details see (https://www.eclipse.org/openj9/)[Eclipse OpenJ9]._

sxa555:
I'm generally ok with the wording but I've put a link to this issue out for circulation to the wider team ;-)

As per an earlier comment from @karianna I would definitely want the 8/11 options marked with `(LTS)` if we're going to include the other versions in the radio box selection. Even if we choose to remove 9/10 in the future I would imagine we'd want 8/11/"Latest" always visible on the front page and so the (LTS) indicator is important

lumpfish:
My concern about the openj9 tip is that it suggests (to me at least) that it would not be suitable for non cloud / container environments, which is certainly not the case.  Maybe something like...

_Eclipse OpenJ9 is the VM that was open sourced by IBM and is also suitable for running all workloads.
This variant uses the same class libraries from OpenJDK, i.e. It is interchangeable from the Hotspot variant as far as Java programmers are concerned.

From more details see (https://www.eclipse.org/openj9/)[Eclipse OpenJ9].

Both VMs have features unique to that VM which may benefit different application scenarios. For more details visit their respective web sites._

sxa555:
I'd probably also make the LTS identifier a link to the support page for people who don't know the term i.e. ([LTS](https://adoptopenjdk.net/support))

sympatheticmoose:
Is there actually enough space within a tooltip to give any true advice on how to choose the best JVM for your particular use case? I feel like this actually requires an overview/comparison piece of content - perhaps a blog on https://blog.adoptopenjdk.net/? Would help drive traffic towards the site. 

The suggested wording to me is just:
1. 'Hotspot is the most popular' 
(I'd also say from the link provided, its not clear exactly where you'd find specific advantages....)
2. 'Eclipse OpenJ9 is basically the same, but its an IBM thing'.

sxa555:
@sympatheticmoose I suspect it might be hard for such an article hosted on adoptopenjdk.net to be written in a suitably objective matter. If anyone clicks the openj9 link from the tooltip content then they will see the advantage claims on the front of the page. At the moment we have an issue where people don't understand because they've often never heard of OpenJ9, and this change makes it more visible on the front page and hopefully get people interested enough to look into and try it. We could, I suppose, link to e.g. https://duckduckgo.com/?q=openj9+vs+hotspot&t=ffab&ia=web for people to look at others' independent research ...

karianna:
Also see https://github.com/AdoptOpenJDK/openjdk-website/issues/301#issuecomment-432932215

sxa555:
@ddunn2 Are you able to make the above changes (Add tooltip with the wording above, add ([LTS](https://adoptopenjdk.net/support)) and put 9 and 10 back in)?

karianna:
I'd prefer not to put 9 and 10 back in as they're no longer secure / supported.

sxa555:
> I'd prefer not to put 9 and 10 back in as they're no longer secure / supported.

That's fair enough for the front page but I'd still prefer to have _some_ way on the website of navigating to the old releases if people really want them

karianna:
> > I'd prefer not to put 9 and 10 back in as they're no longer secure / supported.
> 
> That's fair enough for the front page but I'd still prefer to have _some_ way on the website of navigating to the old releases if people really want them

The archives should cover that.

ddunn2:
@sxa555 I've already started work on it 👍 

Once I have something working I'll push it up for you all to review and see how you feel about the wording along with the look and feel.

ddunn2:
@karianna @sxa555 @SeanC1083 

I've now added the popup to the index page with the text provided above. 

Note this is **NOT** ready to be merged yet, I haven't dealt with medium/small screens just wanted to confirm this is acceptable? 

It's slightly different to the proposed design as there's more text.

sxa555:
@karianna Can you kick another staging refresh (Not sure how that's done)

karianna:
On staging again now.  https://staging.adoptopenjdk.net/339/.  It looks great overall!  A few comments:

1. Putting the Help me choose on a new line reduces the amount of info above the line.  I may just have to completely alter the introduction text as we're clearly struggling here.

2. I had to click on the link for the tooltip to activate.

3. The tooltip shot off to the right and down aways (I'm on a bigger screen)

tooltip

supahgreg:
If the Archive page will actually cover all prior versions, radio inputs might not be the most suitable option.  Once you start getting past a handful of versions (i.e. 8-11) it could get a bit busy.

Also, if the front page omits unsupported releases (which makes sense) it might be worth tweaking the wording to suggest they're still technically available if needed.

```diff
diff --git a/src/handlebars/index.handlebars b/src/handlebars/index.handlebars
index 114004a..3de1554 100644
--- a/src/handlebars/index.handlebars
+++ b/src/handlebars/index.handlebars
@@ -6,7 +6,7 @@
     Java™ is the world's leading programming language and platform. The code for Java is <a href="https://openjdk.java.net/legal/gplv2+ce.html" target="_blank">open source</a> and available at
     <a href="https://openjdk.java.net" target="_blank">OpenJDK</a>™. AdoptOpenJDK provides prebuilt OpenJDK binaries
     from a fully open source set of <a href="https://github.com/AdoptOpenJDK/openjdk-build" target="_blank">build scripts</a> and infrastructure.<br/>
-    Get <a href="https://hub.docker.com/u/adoptopenjdk">Docker Images on Docker Hub</a>. Nightlies can be found in the <a href="./archive.html">Archive</a>.
+    Get <a href="https://hub.docker.com/u/adoptopenjdk">Docker images on Docker Hub</a>, past and present binary releases in the <a href="./archive.html">Archive</a>, and nightlies at <a href="./nightly.html">Nightly builds</a>.
   </p>

   <div class="dl-container">
```

ddunn2:
@karianna 
1. Would you prefer the `Help Me Choose` text to appear to the right of the JVM radio buttons instead? 

2. I'm guessing you want an on hover effect rather than on click?

3. Yeah I haven't added any responsiveness to the popup yet so I imagine it'll look weird on different sized screens. Once we've confirmed we're happy with how it looks I can ensure it works for all screen sizes. 

@supahgreg 
Are you saying you'd prefer the drop down option to remain on the Archive page? Or that we should find a more suitable way to display the versions? 

I'll also update the text to what you've suggested, Thanks for the feedback.

supahgreg:
@ddunn2 Oh, yeah I was only referring to the Archive page.  Keeping the dropdown there would be one option, or perhaps spreading radio inputs across [columns](https://developer.mozilla.org/docs/Web/CSS/columns) would also help as the number of feature releases grows.

karianna:
> @karianna
> 
> 1. Would you prefer the `Help Me Choose` text to appear to the right of the JVM radio buttons instead?
> 2. I'm guessing you want an on hover effect rather than on click?
> 3. Yeah I haven't added any responsiveness to the popup yet so I imagine it'll look weird on different sized screens. Once we've confirmed we're happy with how it looks I can ensure it works for all screen sizes.
> 
> @supahgreg
> Are you saying you'd prefer the drop-down option to remain on the Archive page? Or that we should find a more suitable way to display the versions?
> 
> I'll also update the text to what you've suggested, Thanks for the feedback.

1. Yes, please
2. Yes, please
3. Cool :-)

sxa555:
@karianna One thing we don't have (unless people navigate to the i[nstallation page](https://adoptopenjdk.net/installation.html#) where it's part of the platform dropdown) is a selection of JDK vs JRE. Given that we're doing this switch, what do you think about having a third column of radio buttons with a JDK/JRE selection (defaulting to JDK)

It could alternately be part of the sceond column (Hotspot JDK, Hotspot JRE, OpenJ9 JDK, OpenJ9 JRE) if we wanted to retain horizontal width as-is.

sxa555:
Agree with @karianna that the tooltip should probably pop up above the "Help Me More" instead of below, although perhaps it should also be directly below "Choose a JVM" instead of underneath the options?

Ref @supahgreg's idea of adding an extra line talking about where to get the nightlies I think we need to think a bit more about that - i'm rather reluctant to add much more text to that chunk at the top as it can easily result in people on lower resolution screens having to scroll down if we add more stuff to the top and I think we're at the limit.

ddunn2:
@sxa555 @karianna 

screenshot 2018-10-30 at 09 51 08

This is how the tooltip will look, the reason it's below is due to the difference in screen size (I'm on a MacBook) and I haven't added responsiveness yet incase we wanted anything changing :) 

Following the comments here I'm planning to move the `Help Me Choose` text to the right of the `Choose a JVM` section but keep the popup aligned as you can see in the above screenshot. 

I'll leave the text on the index page unchanged for now until we've settled on what we'd like it to be as @sxa555 makes a good point about the dangers of increasing the size of that text block.

karianna:
@ddunn2 Sounds good, let me know when you've added the commits and I'll trigger a new staging build.

karianna:
> @karianna One thing we don't have (unless people navigate to the i[nstallation page](https://adoptopenjdk.net/installation.html#) where it's part of the platform dropdown) is a selection of JDK vs JRE. Given that we're doing this switch, what do you think about having a third column of radio buttons with a JDK/JRE selection (defaulting to JDK)
> 
> It could alternately be part of the sceond column (Hotspot JDK, Hotspot JRE, OpenJ9 JDK, OpenJ9 JRE) if we wanted to retain horizontal width as-is.

Good point - JREs are still quite popular.  Let's do that in a subsequent PR though.

SeanC1083:
@ddunn2 @karianna @sxa555 I am currently working on some redesigns that include ensuring that users have access to JRE downloads. I will likely share these designs later this week. 

I will put it on the new PR - #331

sxa555:
@ddunn2 Are you able to make the updates?

karianna:
rerunning staging

src/js/index.js
```diff
@@ -15,10 +15,18 @@ function onIndexLoad() {
 /* eslint-enable no-unused-vars */
 
 // INDEX PAGE FUNCTIONS
-
+function removeRadioButtons() {
+  var buttons = document.getElementsByClassName('btn-label');
+  for (var a = 0; a < buttons.length; a++) {
+    if (buttons[a].firstChild.value == 'openjdk9' || buttons[a].firstChild.value == 'openjdk10') {
+      buttons[a].style.display = 'none';
```
tellison: The reason these are hidden is because they are not 'current' releases, i.e. no longer in support at the upstream source projects.
I wonder if it is possible to use an attribute of the builds to determine that when creating the buttons, rather than hard-code the exceptions here?

ddunn2:
@karianna @sxa555 

I've recently pushed a couple of commits. 
The first repositions the popup to the right of the JVM radio button section and changes the popup to appear on hover. 

The second fixes up some minor styling issues for consistency. 

I do have a 3rd commit that I haven't pushed yet, which adds `(LTS)` to all the radio buttons linked to a JDK that has LTS, this was mentioned in this thread previously so if this is also wanted I can push that up. 


Finally... I'm not sure what you want doing for mobile when it comes to the popup? Or even for tablets? For tablets we could possibly enable on click? But for phones I'm not sure what'd be the best experience for a user here. Either remove it or have the popup take over the main device screen?

src/handlebars/index.handlebars
```diff
@@ -13,10 +13,30 @@
...
+      </form>
+      <form id="jvm-selector" class="btn-form">
+        <h3>2. Choose a JVM
+          <span class="help-jvm">Help Me Choose</span>
+          <div class="popup-content" id="jvm-popup">
+            <div class="popup-heading">
+              <h4>Choosing the right JVM</h4>
+            </div>
+            <div class="popup-text">
+              <p class="hotspot-text"><b>Hotspot</b> is the VM from the OpenJDK community. It is the most widely used VM today (e.g. It is used in Oracle's JDK) and is suitable for all workloads.</p>
```
tellison:
```suggestion
              <p class="hotspot-text"><b>Hotspot</b> is the VM from the OpenJDK community. It is the most widely used VM today and is used in Oracle’s JDK. It is suitable for all workloads.</p>
```

src/handlebars/index.handlebars
```diff
@@ -13,10 +13,30 @@
...
+              <p>For more details see <a target="_blank" href="https://openjdk.java.net">OpenJDK</a>.</p>
```
tellison:
```suggestion
              <p>For more details see <a target="_blank" href="https://openjdk.java.net/groups/hotspot/">OpenJDK Hotspot</a>.</p>
```

src/handlebars/index.handlebars
```diff
@@ -13,10 +13,30 @@
...
+              <p class="openj9-text"><b>Eclipse OpenJ9</b> is the VM that was open sourced by IBM and is also suitable for running all workloads. This variant uses the same class libraries from OpenJDK, i.e. It is interchangeable from the Hotspot variant as far as Java programmers are concerned.</p>
```
tellison:
```suggestion
              <p class="openj9-text"><b>OpenJ9</b> is the VM from the Eclipse community.  It is an enterprise-grade VM designed for low memory usage and fast start-up and is used in IBM’s JDK.  It is also suitable for running all workloads.</p>
```

src/handlebars/index.handlebars
```diff
@@ -13,10 +13,30 @@
...
+        <h3>1. Choose a Version</h3>
+      </form>
+      <form id="jvm-selector" class="btn-form">
+        <h3>2. Choose a JVM
+          <span class="help-jvm">Help Me Choose</span>
+          <div class="popup-content" id="jvm-popup">
+            <div class="popup-heading">
+              <h4>Choosing the right JVM</h4>
+            </div>
+            <div class="popup-text">
```
tellison: See below for some suggested PR updates to make the help text more 'symmetrical', and link to the Hotspot VM pages for help.

src/js/index.js
karianna: We could add that to the API - I'll open an issue over there

karianna: https://github.com/AdoptOpenJDK/openjdk-api/issues/118

src/handlebars/index.handlebars
karianna: FWIW - I think they're all good suggestions

src/js/index.js
ddunn2: My latest commit (where the LTS flag is added) actually adds the functionality for us to do this, as we store whether this version is in LTS in the `config.json`. 

So yes it should be possible, I'll push up a proposed fix for you to take a look at. 

Thanks.

ddunn2: Woops... I didn't see your comments @karianna, Would you prefer to leave it in that issue?

ddunn2: I don't have the capability to do it on creation of the buttons yet. I did try that approach but would have needed to refactor various parts of the project. 

I'm about to push up a new commit that will at least remove the `hard coded` version names and will at least future proof against newer JDKs being added in the mean time.

ddunn2:
@supahgreg Thanks for pointing that out, I've just pushed a commit that should resolve that issue :)

karianna:
@ddunn2 I think if you merge @tellisons' changes then we can finally merge this!

karianna:
Hmm, added myself as an assignee to see if I could accept @tellison changes - but still seems to be no option to merge them in.  @ddunn2 - any luck your side?

ddunn2:
@karianna Apologies I was travelling. 

I just want to quickly test this across browsers/devices, I didn't have access to them at the office but I do now! 

Once I've confirmed everything works, and you're happy with it you can merge :) 

Also I was hoping I could merge all of @tellison's changes at once, but looks like I have to do them one at a time :)

ddunn2:
@karianna So there's an issue on tablet/mobile devices. 
Obviously they don't recognise `hover` and adding an `onclick` event for just those devices isn't as simple as I hoped :) 

So there's a few options I can see: 
1. Find a way for users to be able to `onclick` on smaller devices but retain `onhover` for larger screens. 
2. Use `onclick` for devices of all sizes. 
3. Remove the `Help Me Choose` section from smaller devices. 

They're ranked from the most time consuming to the least. I don't mind which approach you decide on, I just wanted to let you know what I've found :).

supahgreg:
Would it be better to simply link to a new page?  Less worry about popup handling, potential for more in-depth content, etc.

karianna:
> @karianna So there's an issue on tablet/mobile devices.
> Obviously they don't recognise `hover` and adding an `onclick` event for just those devices isn't as simple as I hoped :)
> 
> So there's a few options I can see:
> 
> 1. Find a way for users to be able to `onclick` on smaller devices but retain `onhover` for larger screens.
> 2. Use `onclick` for devices of all sizes.
> 3. Remove the `Help Me Choose` section from smaller devices.
> 
> They're ranked from the most time consuming to the least. I don't mind which approach you decide on, I just wanted to let you know what I've found :).

3. For now and lets raise a separate issue for option 1. - That way we can actually land this good work :-)

karianna:
OK - we're finally merging this :-)

Issue #354, opened 2018-11-09
The link for OpenJDK8 (LTS) is pointing to a JRE


d-baer:
The Link on 
https://adoptopenjdk.net/releases.html?variant=openjdk8&jvmVariant=hotspot#x64_win
is pointing to 
https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u192-b12/OpenJDK8U-jre_x64_windows_hotspot_8u192b12.zip

which is a JRE. Seems to be wrong.

karianna:
Interesting - We probably need to resolve this properly in terms of offering JDK and JRE options.

sxa555:
As it happens I just ran into this issue myself :-), so +1

brokenjpl:
Also getting this issue.

A separate link for JRE and the JDK would be very convenient!

karianna:
Resolved for releases.

PR #357, opened 2018-11-10
Rework the Releases page to support Installer+JDK+JRE

supahgreg:
This PR reworks the Releases page to support displaying multiple binary types (e.g Installer, JDK, JRE), resolving #354.

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/releases.js
+++ b/src/js/releases.js
@@ -1,11 +1,7 @@
-var RELEASEDATA;
-
 // When releases page loads, run:
 /* eslint-disable no-unused-vars */
 function onLatestLoad() {
   /* eslint-enable no-unused-vars */
-
-  RELEASEDATA = new Object();
   populateLatest(); // populate the Latest page
 }
 
@@ -31,7 +27,6 @@ function populateLatest() {
 }
 
 function buildLatestHTML(releasesJson) {
-
   // Populate with description
   var variantObject = getVariantObject(variant + '-' + jvmVariant);
   if (variantObject.descriptionLink) {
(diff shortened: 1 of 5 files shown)


supahgreg:
@gdams : Is there anything pending for this, or can it be merged?

karianna:
This is really good/useful - I was cheekily going to ask that you make the same changes for our archive page (probably fairly easy) and home page (harder because of design).  But I'll review and land this instead and make a note in #354 saying that archive and home are left to go (interesting nightly splits them out).

src/js/releases.js
```diff
@@ -40,77 +35,62 @@ function buildLatestHTML(releasesJson) {
...
+    }
+
+    // Skip this asset if it's not a binary type we're interested in displaying
+    const binary_type = releaseAsset.binary_type.toUpperCase();
+    if (['INSTALLER', 'JDK', 'JRE'].indexOf(binary_type) === -1) {
+      return;
+    }
+
+    // Get the existing release asset (passed to the template) or define a new one
+    var release = releases.find((p) => p.platform_name === platform);
```
karianna: nitpick - what's a `p`? :-)

src/js/releases.js
```diff
@@ -40,77 +35,62 @@ function buildLatestHTML(releasesJson) {
...
 
+    // We have the first binary, so add the release asset.
+    if (release.binaries.length === 1) {
+      releases.push(release);
     }
   });
 
-  ASSETARRAY = orderPlatforms(ASSETARRAY);
+  releases = orderPlatforms(releases, 'platform_ordinal');
+  releases.forEach((r) => { r.binaries.sort((a, b) => a.type > b.type ? 1 : a.type < b.type ? -1 : 0) });
```
karianna: `r` -> `release`, I'm also a fan of naming a and b in sort functions for the things they really are but YMMV

supahgreg:
@karianna :  I'm wondering if some of us are working with different NodeJS/NPM versions.  For example, I've been dropping `node_modules` periodically and doing an `npm install` from within the container ([openjdk-website/Dockerfile](https://github.com/AdoptOpenJDK/openjdk-website/blob/master/Dockerfile), which is currently NodeJS 9.x).  Someone else might be working with NodeJS 11.x locally.  Some of the changes are bound to be routine package version tweaks, but some seem a bit odd.  I'll remove the `package-lock.json` change from this since it wasn't essential.

Regarding the other pages: I'm certainly on board with doing something similar there, although as you said perhaps as a later PR.  Mostly just wanted to get this out for review.

karianna:
> @karianna : I'm wondering if some of us are working with different NodeJS/NPM versions. For example, I've been dropping `node_modules` periodically and doing an `npm install` from within the container ([openjdk-website/Dockerfile](https://github.com/AdoptOpenJDK/openjdk-website/blob/master/Dockerfile), which is currently NodeJS 9.x). Someone else might be working with NodeJS 11.x locally. Some of the changes are bound to be routine package version tweaks, but some seem a bit odd. I'll remove the `package-lock.json` change from this since it wasn't essential.
> 
> Regarding the other pages: I'm certainly on board with doing something similar there, although as you said perhaps as a later PR. Mostly just wanted to get this out for review.

Core Node version makes a lot of sense actually!  I wonder if we can lock down the node version we build with (I assume you can state that in package.json somehow?).  Not something to fix now anyhow.  

And thanks for all of the PR's!

src/js/releases.js
supahgreg: Oops-- I had "platform" on the mind, apparently.  I've updated this in ca6da266238294642fbb69949f7d568a4c38ff2c.

src/js/releases.js
supahgreg: Updated in ca6da266238294642fbb69949f7d568a4c38ff2c

PR #360, opened 2018-11-13
Add 'Array.prototype.find' polyfill

supahgreg:
The Releases page redo included a call to [`Array.prototype.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find), which IE doesn't like.  This adds a polyfill to get it working again.

Another option would be to include [`@babel/polyfill`](https://babeljs.io/docs/en/babel-polyfill) to avoid worrying about specific polyfills.

*edit:* I've added a small fix for the Nightly page, since IE also doesn't like [`String.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes).

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/nightly.js
+++ b/src/js/nightly.js
@@ -108,6 +108,7 @@ function buildNightlyHTML(files) {
     var nameOfFile = (eachAsset.binary_name);
     var uppercaseFilename = nameOfFile.toUpperCase(); // make the name of the file uppercase
     NIGHTLYOBJECT.thisPlatform = findPlatform(eachAsset); // get the searchableName, e.g. MAC or X64_LINUX.
+    // We don't use includes because IE doesn't support it well, hence we use indexOf instead
     var type = nameOfFile.indexOf('-jre') !== -1 ? 'jre' : 'jdk';
 
     // secondly, check if the file has the expected file extension for that platform...
--- a/src/js/nightly.js
+++ b/src/js/nightly.js
@@ -108,7 +108,7 @@ function buildNightlyHTML(files) {
     var nameOfFile = (eachAsset.binary_name);
     var uppercaseFilename = nameOfFile.toUpperCase(); // make the name of the file uppercase
     NIGHTLYOBJECT.thisPlatform = findPlatform(eachAsset); // get the searchableName, e.g. MAC or X64_LINUX.
-    var type = nameOfFile.includes('-jre') ? 'jre' : 'jdk';
+    var type = nameOfFile.indexOf('-jre') !== -1 ? 'jre' : 'jdk';
 
     // secondly, check if the file has the expected file extension for that platform...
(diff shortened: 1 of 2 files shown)


src/js/nightly.js
```diff
@@ -108,7 +108,7 @@ function buildNightlyHTML(files) {
     var nameOfFile = (eachAsset.binary_name);
     var uppercaseFilename = nameOfFile.toUpperCase(); // make the name of the file uppercase
     NIGHTLYOBJECT.thisPlatform = findPlatform(eachAsset); // get the searchableName, e.g. MAC or X64_LINUX.
-    var type = nameOfFile.includes('-jre') ? 'jre' : 'jdk';
+    var type = nameOfFile.indexOf('-jre') !== -1 ? 'jre' : 'jdk';
```
karianna: Can you add a comment as to why (e.g. IE is being difficult)

src/js/polyfill/Array.prototype.find.js
```diff
@@ -0,0 +1,46 @@
+// https://tc39.github.io/ecma262/#sec-array.prototype.find
```
karianna: A couple of concerns. 

1. How is this code licensed and do we have explicit permission to reuse it?

2. I'm not sure what this is fixing in terms of a user experience issue

src/js/nightly.js
supahgreg: Do you mean to the code?  The new line is fairly standard JS, so that might not be worthwhile.  The need for this change is due to IE not supporting `includes`, preventing IE users (thinking corporate settings mostly) from using the Nightly page.

karianna: Yeah, I was thinking of adding the comment for IE support, else someone will revert your change down the line...

src/js/polyfill/Array.prototype.find.js
supahgreg: 1. Given the similar comment headers, I believe it's the same origin as as the `findIndex.js` polyfill that is currently in use (MDN).  I can do some checking if needed.
2. The releases page is broken for IE users.

src/js/nightly.js
supahgreg: If that's a concern it might be better to just include a polyfill for `includes`, or the aforementioned `@babel/polyfill` to support everything.

src/js/polyfill/Array.prototype.find.js
supahgreg: Per https://developer.mozilla.org/en-US/docs/MDN/About#Copyrights_and_licenses code samples and snippets added to MDN after 2010 (which the `Array.prototype.find` polyfill was) are in the public domain and require no attribution.

PR #365, opened 2018-11-15
Browserify and Babel polyfill

supahgreg:
This PR, which is currently just here for discussion+review, does (approximately) the minimum to modularize the website's JS, which then enables use of [`@babel/preset-env`](https://babeljs.io/docs/en/babel-preset-env) + [`@babel/polyfill`](https://babeljs.io/docs/en/babel-polyfill) to inject polyfills based upon usage and the `targets` (currently [`defaults`](https://browserl.ist/?q=defaults)).

The main goals of doing this are to:
1. start giving the code a bit more structure
2. enable use of modern JS without worrying as much about browser compatibility

##### Checklist

- [x] `npm test` passes


src/js/app.js
```diff
@@ -0,0 +1,5 @@
+global.onIndexLoad = require('./index').onIndexLoad;
```
karianna: I thought this file got generated?

supahgreg: It does-- I just unintentionally chose the same name for the entry point (`src/js/app.js`) and the final result (`dist/js/app.js` and `dist/js/app.min-*.js`).  Either could be changed to something else (e.g. `src/js/entry.js`) to make things unique.

karianna: To avoid confusion I think that would be wise.

supahgreg: Okay, I've renamed it to `entry.js` (924b2409e12162808d2e010ba5fe6241eb7d9264).

supahgreg:
I went ahead and did a bit in 51c4c2db5e9625b40a8be017284aa63cd89f97ec.  Most of that commit is starting to clean up what's currently in use, as opposed to making changes to how things work.

karianna:
@supahgreg Merge conflicts.

karianna:
OpenJDK8 and OpenJ9 combination fails on staging home page: https://staging.adoptopenjdk.net/365/?variant=openjdk8&jvmVariant=openj9

Everything else seems to be OK.

.eslintrc.yml
```diff
@@ -8,29 +8,6 @@ rules:
   no-inner-declarations: 0
   quotes: [2, single]
 globals:
-  moment: true
```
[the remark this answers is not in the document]
supahgreg: Previously ESLint was told to ignore many things that existed in the global scope at runtime but weren't defined within the context of a single JS script.  For example: `src/js/archive.js` used `loadPlatformsThenData` from `0-global.js`.

By requiring explicit imports (e.g. `const loadPlatformsThenData = require('./common');`) these things are no longer in the global scope, so we can let ESLint work normally.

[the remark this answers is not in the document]
karianna: Awesome, thanks for the explanation!

karianna:
> OpenJDK8 and OpenJ9 combination fails on staging home page: https://staging.adoptopenjdk.net/365/?variant=openjdk8&jvmVariant=openj9
> 
> Everything else seems to be OK.

I think just this issue to look at/fix and we can land this, thank you!

supahgreg:
Oops-- should be good now.

PR #373, opened 2018-11-23
[WIP] Add support for website to detect macos large heap builds

sxa555:
- [ ] `npm test` passes

WIP since I'm not sure what I'm doing, but first step in trying to make macos OpenJ9 large heap builds show as such on the web site


karianna:
I don't see it on the staging site - so perhaps a bug somewhere?  You can actually run the site locally BTW. `npm start`. It talks to the PRD API etc, so you can then debug locally.

sxa555:
Yeah it hasn't worked. I wanted to get some code up to see if the "quick fix" would work but it hasn't so will debug locally now

sxa555:
Can't see what I'm missing here - the builds are still being detected and labelled as the normal heap versions, but I seen to have the same changes as for Linux, unless it's related to the way the array is sorted adjusting the precedence when a filename matches two `searchableName`s ... May be one to leave for next week ...

karianna:
> Can't see what I'm missing here - the builds are still being detected and labelled as the normal heap versions, but I seen to have the same changes as for Linux, unless it's related to the way the array is sorted adjusting the precedence when a filename matches two `searchableName`s ... May be one to leave for next week ...

Can pair on it next week - I solved a similar issue a few weeks ago.

sxa555:
Table is populated from html which comes from `buildNightlyHTML` in `src/js/nightly.js`. The `findPlatform()` function in `src/js/common.js` retrieves the searchable name and that's not giving back the `MACOSXL`

karianna:
@sxa555 Bug was likely https://github.com/AdoptOpenJDK/openjdk-api/issues/134 - this is resolved so I think you can try again :-)

sxa555:
Looks good now - removing WIP

Issue #417, opened 2019-03-06
Add openjdk12 to nightly builds download page


lumpfish:
Now we have nightly builds for jdk12 users should be able to get at them via https://adoptopenjdk.net/nightly.html.

At the moment there is no 'OpenJDK 12' option.

PR #418, opened 2019-03-07
add jdk12 support to the website

gdams:
closes: https://github.com/AdoptOpenJDK/openjdk-website/issues/417

##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/common.js
+++ b/src/js/common.js
@@ -262,7 +262,13 @@ module.exports.setRadioSelectors = () => {
       btnLabel.appendChild(input);
 
       if (group === 'jdk') {
-        btnLabel.innerHTML += `<span>${variant.label}${variant.lts ? ' (LTS)' : ''}</span>`;
+        if (variant.lts === true){
+          btnLabel.innerHTML += `<span>${variant.label} (LTS)</span>`;
+        } else if (variant.lts === 'latest') {
+          btnLabel.innerHTML += `<span>${variant.label} (Latest)</span>`;
+        } else {
+          btnLabel.innerHTML += `<span>${variant.label}</span>`;
+        }
       } else {
         btnLabel.innerHTML += `<span>${variant.jvm}</span>`;
       }
--- a/src/json/config.json
+++ b/src/json/config.json
@@ -64,6 +64,22 @@
       "label": "OpenJDK 11",
       "lts": true,
       "descriptionLink": "https://www.eclipse.org/openj9"
+    },
+    {
+      "searchableName": "openjdk12-hotspot",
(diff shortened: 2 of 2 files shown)


SueChaplain:
@gdams - I don't see 12 when I come to this page: https://staging.adoptopenjdk.net/418/ .. but I do see it on the latest release, nightlys, and archive page.

gdams:
It's because it's not an LTS release. We only show LTS releases on the website. We could perhaps add the option to show `latest`. CC @karianna ?

SueChaplain:
Looking good! It would be good to lose OpenJDK 9 & 10 from all but the archive page (think I'm correct in that these are no longer built) but I guess that is for another issue.

karianna:
> Looking good! It would be good to lose OpenJDK 9 & 10 from all but the archive page (think I'm correct in that these are no longer built) but I guess that is for another issue.

I think it's OK to have the 9 and 10 builds for historical spelunkers.  But it's a good Q for the TSC, I'll dd a label so we pick it up.

karianna:
> It's because it's not an LTS release. We only show LTS releases on the website. We could perhaps add the option to show `latest`. CC @karianna ?

TSC decision I think - adding label.

gdams:
TSC has voted to not show JDK 12 on home page until first official Short term release has been shipped

ben-walsh:
I understand the reasoning behind the decision for the home page. However, can we get the Nightly part of this completed ? It is always good to start getting feedback from the "early adopters" as soon as possible.

karianna:
> I understand the reasoning behind the decision for the home page. However, can we get the Nightly part of this completed? It is always good to start getting feedback from the "early adopters" as soon as possible.

I'd be OK with that 0 this PR would need to change though.

SueChaplain:
Hi @gdams or @karianna  -  I'm preparing some "What's new" content for the OpenJ9 0.13.0 release that will include a link to https://adoptopenjdk.net/releases.html?variant=openjdk12&jvmVariant=openj9. When do you expect this PR to merge please? I can put up "coming soon" and modify later to avoid a broken link, but would be good if I can avoid the extra step.

karianna:
Assigned to @gdams - hopefully he can get some time to adjust this PR (but he's certainly loaded up today I know!)

gdams:
@SueChaplain nightly builds are already available here https://adoptopenjdk.net/nightly.html?variant=openjdk12&jvmVariant=openj9. Once we have an official release we will merge this PR.

SueChaplain:
@gdams - ok thanks .... will hang on for the main release page.

Issue #422, opened 2019-03-15
modify installation page for windows to simplify hash verification


userquin:
Instead compare visually the hash, just provide a command to verify it (replace `XXXX` with the exe file and `XXXX_HASH` with the SHA256 hash):

```shell
certutil.exe -hashfile XXXX.exe SHA256 2>&1 | find /I "XXXX_HASH">Nul && ( echo "XXXX.exe: The SHA-256 fingerprint matches" ) || ( echo "XXXX.exe: The SHA-256 fingerprint does NOT match" )
```

spanish translation:

```shell
certutil.exe -hashfile XXXX.exe SHA256 2>&1 | find /I "XXXX_HASH">Nul && ( echo "XXXX.exe: La huella SHA-256 coincide" ) || ( echo "XXXX.exe: La huella SHA-256 NO coincide" )
```

userquin:
When selecting windows JRE/JDK on https://adoptopenjdk.net/installation.html

karianna:
Looks good, Pull Request welcome!

userquin:
@karianna I'll take a look to make a PR...

userquin:
@karianna It is posible to request server api to returns the sha256 on the response? If no, then we need to resolve the url and parse the text async 

If the answer is NO,  can I use Promises on the project?:

```javascript
module.exports.loadChecksum = url => {
  return new Promise((resolve, reject) => {
    loadUrl(url, (response) => {
      try {
        resolve(toJson(response));
      } catch (e) {
        reject(e);
      }
    });
  });
}
```

```json
```
![imagen](https://user-images.githubusercontent.com/6311119/54464573-82a45980-4777-11e9-9805-c1398c1e17d5.png)

userquin:
I have a problem with cors requesting sha256.txt file to github, how can I fix it? (installing gulp-connect and add server:test task): i'm not an expert on gulp, normally I work woth webpack:

![imagen](https://user-images.githubusercontent.com/6311119/54468265-21868100-478b-11e9-86c9-dd0138896585.png)

userquin:
I have upload some changes to https://github.com/userquin/openjdk-website:
- config.schema.json: added props for auto command and auto result command (as not required props)
- common.js: added logic to get these 2 props and retrieve sha256 text file from server (parsing it from xhr blob response)
- common.js: modification on loadUrl to allow get blob instead responseText
[...]

PR #425, opened 2019-03-17
PR for 422

userquin:
Due to the fact client cannot download sha256.txt file (see https://github.com/AdoptOpenJDK/openjdk-website/issues/422#issuecomment-473655077), the changes are not finished, just need to parse the file and replace FILEHASH on the `autoResultCommand` text.

### DIFF
--- a/src/js/common.js
+++ b/src/js/common.js
@@ -80,11 +80,11 @@ module.exports.getInstallCommand = (searchableName) => lookup[searchableName].in
 // gets the CHECKSUM COMMAND when you pass in 'searchableName'
 module.exports.getChecksumCommand = (searchableName) => lookup[searchableName].checksumCommand;
 
+// gets the CHECKSUM AUTO COMMAND HINT when you pass in 'searchableName'
+module.exports.getChecksumAutoCommandHint = (searchableName) => lookup[searchableName].checksumAutoCommandHint;
+
 // gets the CHECKSUM AUTO COMMAND when you pass in 'searchableName'
 module.exports.getChecksumAutoCommand = (searchableName) => lookup[searchableName].checksumAutoCommand;
-
-// gets the CHECKSUM AUTO COMMAND RESULT when you pass in 'searchableName'
-module.exports.getChecksumAutoResultCommand = (searchableName) => lookup[searchableName].checksumAutoResultCommand;
 
 // gets the PATH COMMAND when you pass in 'searchableName'
 module.exports.getPathCommand = (searchableName) => lookup[searchableName].pathCommand;
@@ -160,22 +160,6 @@ module.exports.loadLatestAssets = (variant, openjdkImp, releaseType, release, ty
 
   const url = `https://api.adoptopenjdk.net/v2/latestAssets/${releaseType}/${variant}`;
   queryAPI(release, url, openjdkImp, type, errorHandler, handleResponse);
-}
-
-module.exports.getChecksum = url => checksums[url]
-
-module.exports.loadChecksum = (url, message) => {
-  return new Promise((resolve, reject) => {
-    loadUrl(url, sha256Content => {
-      try {
-        const checksum = sha256Content.split(' ')[0].trim();
-        checksums[url] = checksum;
-        resolve(message.replace('FILEHASH', checksum));
-      } catch (e) {
-        reject(e);
-      }
-    });
-  });
 }
 
 function loadUrl(url, callback) {
--- a/src/js/common.js
+++ b/src/js/common.js
@@ -7,6 +7,9 @@ const {platforms, variants} = require('../json/config');
 const lookup = {};
 platforms.forEach((platform) => lookup[platform.searchableName] = platform);
 
+// checksums cache
+const checksums = {};
+
 let variant = module.exports.variant = getQueryByName('variant') || 'openjdk8';
 let jvmVariant = module.exports.jvmVariant = getQueryByName('jvmVariant') || 'hotspot';
 
@@ -76,6 +79,12 @@ module.exports.getInstallCommand = (searchableName) => lookup[searchableName].in
 
 // gets the CHECKSUM COMMAND when you pass in 'searchableName'
 module.exports.getChecksumCommand = (searchableName) => lookup[searchableName].checksumCommand;
+
+// gets the CHECKSUM AUTO COMMAND when you pass in 'searchableName'
+module.exports.getChecksumAutoCommand = (searchableName) => lookup[searchableName].checksumAutoCommand;
+
+// gets the CHECKSUM AUTO COMMAND RESULT when you pass in 'searchableName'
+module.exports.getChecksumAutoResultCommand = (searchableName) => lookup[searchableName].checksumAutoResultCommand;
 
 // gets the PATH COMMAND when you pass in 'searchableName'
 module.exports.getPathCommand = (searchableName) => lookup[searchableName].pathCommand;
@@ -153,12 +162,43 @@ module.exports.loadLatestAssets = (variant, openjdkImp, releaseType, release, ty
   queryAPI(release, url, openjdkImp, type, errorHandler, handleResponse);
 }
 
+module.exports.getChecksum = url => checksums[url]
+
+module.exports.loadChecksum = (url, message) => {
+  return new Promise((resolve, reject) => {
+    loadUrl(url, response => {
+      try {
+        const reader = new FileReader();
+        // This fires after the blob has been read/loaded.
+        reader.addEventListener('loadend', e => {
+          // todo@userquin: fix this -> https://developer.mozilla.org/en-US/docs/Web/API/Event/srcElement
+          console.info(e.srcElement);
+          console.info(e.target);
+          const text = e.srcElement.result;
+          const checksum = text.split(' ')[0].trim();
+          console.info(checksum);
+          checksums[url] = checksum;
+          resolve(message.replace('FILEHASH',checksum));
+        });
+        reader.readAsText(response.blob());
+      } catch (e) {
+        reject(e);
+      }
+    }, false);
+  });
+}
+
 function loadUrl(url, callback) {
+  const checksumRequest = url.endsWith('sha256.txt');
   const xobj = new XMLHttpRequest();
   xobj.open('GET', url, true);
   xobj.onreadystatechange = () => {
     if (xobj.readyState == 4 && xobj.status == '200') { // if the status is 'ok', run the callback function that has been passed in.
-      callback(xobj.responseText);
+      if (checksumRequest) {
+        callback(xobj);
+      } else {
+        callback(xobj.responseText);
+      }
     } else if (
       xobj.status != '200' && // if the status is NOT 'ok', remove the loading dots, and display an error:
       xobj.status != '0') { // for IE a cross domain request has status 0, we're going to execute this block fist, than the above as well.
--- a/src/js/common.js
+++ b/src/js/common.js
@@ -166,39 +166,24 @@ module.exports.getChecksum = url => checksums[url]
 
 module.exports.loadChecksum = (url, message) => {
   return new Promise((resolve, reject) => {
-    loadUrl(url, response => {
+    loadUrl(url, sha256Content => {
       try {
-        const reader = new FileReader();
-        // This fires after the blob has been read/loaded.
-        reader.addEventListener('loadend', e => {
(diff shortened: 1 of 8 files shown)


gulpfile.js
```diff
@@ -212,6 +212,7 @@ gulp.task('browser-sync', () => {
         server: {
             baseDir: './'
         },
+        cors: true,
```
karianna: Security concern?

src/js/common.js
```diff
@@ -153,12 +162,43 @@ module.exports.loadLatestAssets = (variant, openjdkImp, releaseType, release, ty
...
+
+module.exports.loadChecksum = (url, message) => {
+  return new Promise((resolve, reject) => {
+    loadUrl(url, response => {
+      try {
+        const reader = new FileReader();
+        // This fires after the blob has been read/loaded.
+        reader.addEventListener('loadend', e => {
+          // todo@userquin: fix this -> https://developer.mozilla.org/en-US/docs/Web/API/Event/srcElement
+          console.info(e.srcElement);
```
karianna: Would need to remove the console logging when done

src/js/common.js
```diff
@@ -153,12 +162,43 @@ module.exports.loadLatestAssets = (variant, openjdkImp, releaseType, release, ty
...
 
+module.exports.getChecksum = url => checksums[url]
+
+module.exports.loadChecksum = (url, message) => {
+  return new Promise((resolve, reject) => {
+    loadUrl(url, response => {
+      try {
+        const reader = new FileReader();
+        // This fires after the blob has been read/loaded.
+        reader.addEventListener('loadend', e => {
```
karianna: What's an `e`?

src/js/common.js
```diff
@@ -153,12 +162,43 @@ module.exports.loadLatestAssets = (variant, openjdkImp, releaseType, release, ty
...
+          console.info(e.target);
+          const text = e.srcElement.result;
+          const checksum = text.split(' ')[0].trim();
+          console.info(checksum);
+          checksums[url] = checksum;
+          resolve(message.replace('FILEHASH',checksum));
+        });
+        reader.readAsText(response.blob());
+      } catch (e) {
+        reject(e);
```
karianna: What happens to the user experience on a rejection?

gulpfile.js
userquin: This is to allow access 302 redirection; if not included when connecting to download sha256.txt file, then the broswer will reject the request
On the other hand, if the response to the redirection from github does not includes cors headers on 302 response, the browser also reject the request

src/js/common.js
userquin: yes, once tested this console will be removed

src/js/common.js
userquin: e is the event once the blob is resolved: like another addEventListener callback

src/js/common.js
userquin: in taht case the div container will not be shown, initially is hidden

userquin:
> Made some comments but one of the more experienced web devs needs to chime in here. Also wondering if this is universally beneficial given its only windows.

My intention is to made it available to all platforms:

For linux/unix:
```shell
echo "FILEHASH  FILENAME" 2>&1 | sha256sum -c
```

and for macosx:
```shell
echo "FILEHASH FILENAME" 2>&1 | shasum -a 256 -c
```

src/json/config.schema.json
```diff
@@ -142,6 +142,18 @@
             "title": "Checksum command",
             "examples": ["sha256sum FILENAME"]
           },
+          "checksumActionCommand": {
+            "$id": "#/properties/platforms/items/properties/checksumActionCommand",
+            "$ref": "#/definitions/command",
```
userquin: the `#id` should be `#/properties/platforms/items/properties/checksumAutoCommand`

src/json/config.schema.json
```diff
@@ -142,6 +142,18 @@
...
+            "title": "Checksum auto command",
+            "examples": ["sha256sum FILENAME"]
+          },
+          "checksumActionCommandResult": {
+            "$id": "#/properties/platforms/items/properties/checksumActionResultCommand",
+            "type": "string",
```
userquin: the `#id` should be `#/properties/platforms/items/properties/checksumAutoResultCommand`

src/js/common.js
userquin: In fact, I must check if responseText from xhr redirect response just returns the checksum line: in that case we just revert all changes made to handle xhr response on `commons.js` and here just remove the FileReader code...

src/js/common.js
userquin: just see the staging server, there are errors downloading all sha256.txt files, and the div container is hidden, the page is the same as the original

userquin:
@karianna the optimael solution will be if the server api includes the hash for each platform, this will prevent download each sha256.txt file and the auto comannd can be directly calculated.

The code is ready to handle this situation, and if the server api is changed to oncluded the hash, then 95% of the changes can be removed.

gdams:
>  if the server api is changed to include the hash, then 95% of the changes can be removed.
@userquin I'm kind of pro doing work on the API to render the hash as a string and return it.

userquin:
@gdams It is posible to include for each platform the checksum + checksum hash in the server api?

gdams:
> @gdams It is posible to include for each platform the checksum + checksum hash algorithm in the server api response?

Yeah I can certainly take a look at that @userquin

userquin:
@gdams any news on this?

karianna:
@userquin Merge conflicts

userquin:
This PR is not ready to be merged yet, I'm still waiting for @gdams responses. As I said, if the server api comes with hash and hash algh, we need only to adapt the ui, but if we need to call server to download each hash file, then some things must be changed on server side to avoid CORS errors.

userquin:
Ok, it seems that client can now download hash files, so we need only update the client to read its content.

userquin:
I'm at office, later at home I'll check it again, I'll try to add logic to parse the response, seems to be failing for same reason.

userquin:
To solve the conflict (I dont know from what branch):
For current 2.0.1:
```javascript
// browser-sync task
gulp.task('browser-sync', function() {
    browserSync.init({
        server: {
            baseDir: './'
        },
        cors: true,
        notify: false
    });
});

// robots task
gulp.task('robots', function () {
  gulp.src('index.html')
    .pipe(robots({
      useragent: '*',
      allow: ['/'],
      disallow: ['/404.html', '/banner.html', '/dist']
    }))
    .pipe(gulp.dest('./'));
});
```
or for this conflict:
```javascript
// browser-sync task
gulp.task('browser-sync', function() {
    browserSync.init({
        server: {
            baseDir: './'
        },
        cors: true,
        notify: false
    });
});

// robots task
gulp.task('robots', function () {
  gulp.src('index.html')
    .pipe(robots({
      useragent: '*',
      allow: ['/'],
      disallow: ['/404.html', '/dist']
    }))
    .pipe(gulp.dest('./'));
});
```

userquin:
@karianna The response for api data now has the cors header, but we also need cors header on responses for sha256.txt files:

api_call

sha256 txt

cors-errors

userquin:
Well, by passing cors on chrome, just works, I need to do some final checks and I will update the PR. I need to check it on staging, but will not work due to CORS headers...

userquin:
```javascript
2637dab3bc81274e19991eebc27684276b482dd71d0f84fedf703d4fba3576e5  OpenJDK8U-jdk_x64_windows_hotspot_8u202b08.zip
```
cors_disabled_chrome2

cors_disabled_chrome

userquin:
@karianna I just push to my repo the changes and now it is working (see link below), and updated mac os x commands, I cannot test on linux now, maybe tomorrow  (it's 5 o'clock in the morning and I'm going to sleep;))..

I have tested on local on a Windows 10, starting chrome with no security: https://alfilatov.com/posts/run-chrome-without-cors/

src/js/common.js
userquin: there is no need to use FileReader, just xhr.responseText works

@karianna to listen for events on html elements you add event listeners to them, provinding the event in question and a callback function: for example, a button can emit multiple events such as click, then we register an event listener to be notified when the user clicks on it:
```javascript
someButton.addEventListener("click", function(e) {
  // handle the event
}
```
The callback function can receive the event that fires the callback, you can call what the name you want. Using arrow functions, we can reduce the code removing extra code that is implicit:
```javascript
someButton.addEventListener("click", e => {
  // handle the event
}
```

userquin:
Upps, there is something wrong: before update the master I just update but resolving the conflict updates changes @gdams do:

imagen

Some files was changed, but not present on my master branch (package.json and gulp.js), and at least gulp.js is wrong on my branch and in this PR (robots missing and robots task also missing on my gulp.js).

I dont want to push my changes, waiting to see where we are and how to solve the problem.

userquin:
@gdams where are you touching? Your 2 last changes are missing from this master and I dont know where are included. See previous image.

userquin:
@gdams Once I resolve the conflict manually on gulp.js file, I have compared my updated version againts your changes: as you can see bellow on the images, my gulp file has references to robots but clipboard no (copied from this master)

imagen

imagen

userquin:
I'm trying another approach to avoid download the sha256.txt files for each platform, just using curl or wget  to download sha256 file and then check it, it is a good solution with only one command (for example, for mac):
```shell
curl -O -L FILEHASHURL && shasum -a 256 -c FILEHASHNAME
```
We can also hide download progress including `-s` flag:
```shell
curl -O -L -s FILEHASHURL && shasum -a 256 -c FILEHASHNAME
```
mac_checksum_alternative_command

userquin:
For linux/unix (tested only on ubuntu 18):
```shell
wget -O- -T 1 -t 1 FILEHASHURL  | sha256sum -c
```
We can also hide download progress including `-q` flag:
```shell
wget -O- -q -T 1 -t 1 FILEHASHURL  | sha256sum -c
```
Captura de pantalla de 2019-04-18 16-21-40

userquin:
And finally, after some tedious work, for windows (only `PowerShell >= 3.0`):

```shell
powershell -command "[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12;  iwr -outf FILEHASHNAME FILEHASHURL" && powershell "$CHECKSUMVAR=($(Get-FileHash -Algorithm SHA256 -LiteralPath FILENAME | Format-List -Property Hash | Out-String) -replace \"Hash : \", \"\" -replace \"`r`n\", \"\"); Select-String -LiteralPath FILEHASHNAME -Pattern $CHECKSUMVAR | Format-List -Property FileName | Out-String" | find /i "FILEHASHNAME">Nul && ( echo "FILENAME: The SHA-256 fingerprint matches" ) || ( echo "FILENAME: The SHA-256 fingerprint does NOT match" )
```
windows_checksum_alternative_command

userquin:
With this new approach, there is no need to download the sha256 files, so when the user executes the command (for ist corresponding SO), the sha256 file is downloaded and the file verified.
I'm updating the ui to reflect this changes.

userquin:
@karianna, @gdams For windows I have not found a way to do it using windows shell, if some expert on it can rewrite the command instead using PowerShell, just need to replace the command on config.json file.

karianna:
@gdams I'll await your review next.

userquin:
@gdams can you ignore `package.json` and `package-lock.json`  from my PR? I dont know who and when touched these files and in which branch. I only updated my master branch  followed by a  `npm install` before making and pushing last changes in my master, and the result were those `package.json` and `package-lock.json` files.

karianna:
@userquin I think you'll need to rebase/resolve the conflict and push again then I think this is good to go.

userquin:
@karianna later at home I'll try or maybe tomorrow

userquin:
@karianna there is no way to rebase package.json and package-lock.json from upstream/master, I'll try to just copy/paste both files into my master and try again...

userquin:
ok, I just forget to do `npm install`, but then all versions change on package-lock.json and some dependencies were removed  (see images below).

just push or wait?

imagen

imagen
imagen

karianna:
> ok, I just forget to do `npm install`, but then all versions change on package-lock.json and some dependencies were removed (see images below).
> 
> just push or wait?
> 
> imagen
> 
> imagen
> imagen

Yes push.

userquin:
browser-sync fails on travis ci build, it needs browser-sync-client module to run, so I need to add it to `package.json`:
```json
    "browser-sync-client": "^2.26.4", <=== to fix build
    "browser-sync": "^2.26.5",
```

PR #428, opened 2019-03-20
render checksums in browser instead of requiring a download

gdams:
##### Checklist

- [ ] `npm test` passes

### DIFF
--- a/src/js/common.js
+++ b/src/js/common.js
@@ -314,3 +314,15 @@ module.exports.setRadioSelectors = () => {
     }
   }
 }
+
+global.renderChecksum = function(link) {
+  var settings = {
+    "url": `https://cors-anywhere.herokuapp.com/${link}`,
+    "method": "GET",
+  }
+  $.ajax(settings).done(function(response) {
+    var checksum = response.split(' ')[0];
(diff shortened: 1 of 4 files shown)


karianna:
I'm not sure if the value of this outweighs the weight of bringing in two more libs.

src/js/common.js
```diff
@@ -314,3 +314,15 @@ module.exports.setRadioSelectors = () => {
     }
   }
 }
+
+global.renderChecksum = function(link) {
+  var settings = {
+    "url": `https://cors-anywhere.herokuapp.com/${link}`,
```
BethGriggs: From https://github.com/Rob--W/cors-anywhere/
> A public demo of CORS Anywhere is available at https://cors-anywhere.herokuapp.com. This server is only provided so that you can easily and quickly try out CORS Anywhere

gdams: thanks @BethGriggs I'm going to have to get it to work with cors anyway so will need to investigate fixing the original issue and not hacking with https://cors-anywhere.herokuapp.com.

gdams:
closing.

PR #472, opened 2019-04-26
Change 8u official release to 8u212-b04

jerboaa:
That build includes an important performance regression fix.


jerboaa:
Thanks, Tim!

PR #496, opened 2019-05-22
Add JRE tarballs to upstream.html

jerboaa:
This also restructures the table a bit with updated styling so
it's easier to navigate with these added options.


jerboaa:
> Maybe have aarch64 as a separate column? Hard to read otherwise

Thanks, I'll look into it.

FWIW, the way forward would be to have support for this in the API and this table mess goes away entirely.

jerboaa:
Here is another attempt. This looks cleaner, I believe. Thoughts?

PR #520, opened 2019-06-27
Fix incorrect file names showing up in copied installation commands

### LINKED ISSUE #519: Hash verificaction wrong
@karianna, @gdams 

From this link https://adoptopenjdk.net/installation.html#x64_win-jre when copying the verification command using copy button the hash file to calculate the hash has the wrong target, that is, the hash file for the command is for JDK not for JRE:

If you see the command, it is OK:
`certutil -hashfile OpenJDK8U-jre_x64_windows_hotspot_8u212b04.zip SHA256`
but using the copy button:
`certutil -hashfile OpenJDK8U-jdk_x64_windows_hotspot_8u212b04.zip SHA256`

I don't know if also wrong on other targets (linux, mac, etc...).

Maybe the rest api is serving wrong data.

ParkerM:
Fixes #519, tested in IE11 and Firefox 67.

The existing method grabs text from the DOM using an injected ID param based on platform (e.g. `#java-X64_WIN`) and does not account for JDK vs JRE selection.

This change uses the more-generic `previousSibling` selector to grab the text content of its nearest preceding sibling.

##### Checklist
- [x] `npm test` passes

### DIFF
--- a/src/js/installation.js
+++ b/src/js/installation.js
@@ -7,9 +7,10 @@ const loading = document.getElementById('loading');
 const errorContainer = document.getElementById('error-container');
 const platformSelector = document.getElementById('platform-selector');
 
-global.copyClipboard = (elementSelector) => {
+global.copyPreviousSibling = (target) => {
+  const cmd = target.previousSibling.textContent;
   const input = document.createElement('input');
-  input.value = document.querySelector(elementSelector).textContent;
+  input.value = cmd;
 
   document.body.appendChild(input);
   input.select();
@@ -18,7 +19,7 @@ global.copyClipboard = (elementSelector) => {
   alert('Copied to clipboard');
 
   document.body.removeChild(input);
-}
+};
 
 module.exports.load = () => {
   setRadioSelectors();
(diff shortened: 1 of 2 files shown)


src/js/installation.js
```diff
@@ -7,9 +7,10 @@ const loading = document.getElementById('loading');
 const errorContainer = document.getElementById('error-container');
 const platformSelector = document.getElementById('platform-selector');
 
-global.copyClipboard = (elementSelector) => {
+global.copyPreviousSibling = (target) => {
```
karianna: What are we actually copying, can we name the function for what it is copying?

src/js/installation.js
```diff
@@ -7,9 +7,10 @@ const loading = document.getElementById('loading');
...
+  const cmd = target.previousSibling.textContent;
   const input = document.createElement('input');
-  input.value = document.querySelector(elementSelector).textContent;
+  input.value = cmd;
```
karianna: `cmd` -> `checksumCmd`?

src/js/installation.js
ParkerM: Agreed, will push a change here shortly.

PR #564, opened 2019-09-13
Mention "on their infrastructure" for upstream builds

jerboaa:
This should make it clearer that upstream builds are being built on Red Hat infrastructure (not AdoptOpenJDK's infrastructure).


jerboaa:
Change visible on staging here: https://staging.adoptopenjdk.net/564/upstream.html

PR #567, opened 2019-09-14
Added LTS definition in the Help Me Choose index popup-text

cesarhernandezgt:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/handlebars/index.handlebars
+++ b/src/handlebars/index.handlebars
@@ -34,6 +34,7 @@
               <p>For more details see <a target="_blank" href="https://openjdk.java.net/groups/hotspot/">OpenJDK HotSpot</a>.</p>
               <p class="openj9-text"><b>Eclipse OpenJ9</b> is the VM from the Eclipse community.  It is an enterprise-grade VM designed for low memory usage and fast start-up and is used in IBM’s JDK.  It is suitable for running all workloads.</p>
               <p>For more details see <a target="_blank" href="https://www.eclipse.org/openj9/">Eclipse OpenJ9</a>.</p>
+              <p class="openj9-text"><b>Long Term Support (LTS)</b> in OpenJDK is really just an understanding between the various contributors (led by Oracle and Red Hat) that the code line for Java SE 11 / 17 / 23 etc will be maintained for a longer period of time than six months.</p>
             </div>
           </div>
         </h3>


ggam:
Great idea, but I thought Java 14 and 17 would be "LTS" too? Sadly I can't find the source for any of the two statements. Do you have any link?

Given all the confusion there's around the LTS stuff, it might be useful to even add a specific page with a more complete explanation. What do you think?

cesarhernandezgt:
Hi @ggam,
Thanks for the feedback, my original intent was to just provide to end user the deffinition of LTS, I copied the text from the JC [Java is still free](https://medium.com/@javachampions/java-is-still-free-2-0-0-6b9aa8d6d244).

Now that I see your feedback I think we can actually have two paths:

**1. Add a link to LTS as you suggested in order to provide the 14,17, etc. release cadence** 

Pros: 

- We provide more information to end-user.

Cons: 

- The 14,17 will need to be maintained in the future along with the link.
- The "one" link that explains all about LTS is hard to find, notice that even [Oracle](https://www.oracle.com/technetwork/java/java-se-support-roadmap.html) nor [Wikipedia](https://en.wikipedia.org/wiki/Java_version_history) don't provide information yet for future releases beyond 13.

**2. Simplify the PR by removing the version numbers, a new proposal could be:**
`
Long Term Support (LTS) in OpenJDK is really just an understanding between the various contributors (led by Oracle and Red Hat) that the code line for a specific version of Java SE  will be maintained for a longer period of time than six months.
`

Pros: 

- Removing the version numbers avoid maintenance and provide enough information for understanding LTS definition.

Cons:
- Is up to the end-user to investigate by their own the latest Oracle release cadence and LTS support roadmap.

I vote for option #2. 
What do you think about it?

ggam:
Option 2 with a link to Java is Still Free article for more information sounds pretty reasonable!

My only concern about the text is that, to my knowledge, OpenJDK by itself doesn't have the LTS notion, meaning *anyone* can provide LTS for *any* specific JDK version. If that's not mistaken, then it's not true that "the contributors, led by Oracle and Red Hat" decide which version has long term support.

I might be wrong on this and OpenJDK might indeed support the LTS notion in the form of the "updates" projects. But again, any company could just maintain their own private repo and provide support.

This is just semantic stuff, but given all the confusion it has created, we need to be careful and correctly edudate users.

src/handlebars/index.handlebars
```diff
@@ -34,6 +34,7 @@
               <p>For more details see <a target="_blank" href="https://openjdk.java.net/groups/hotspot/">OpenJDK HotSpot</a>.</p>
               <p class="openj9-text"><b>Eclipse OpenJ9</b> is the VM from the Eclipse community.  It is an enterprise-grade VM designed for low memory usage and fast start-up and is used in IBM’s JDK.  It is suitable for running all workloads.</p>
               <p>For more details see <a target="_blank" href="https://www.eclipse.org/openj9/">Eclipse OpenJ9</a>.</p>
+              <p class="openj9-text"><b>Long Term Support (LTS)</b> in OpenJDK is really just an understanding between the various contributors (led by Oracle and Red Hat) that the code line for Java SE 11 / 17 / 23 etc will be maintained for a longer period of time than six months.</p>
```
karianna: class openj9-text should also be removed.

gdams:
closing in favour of https://github.com/AdoptOpenJDK/openjdk-website/pull/608

PR #657, opened 2020-01-15
Add banner to notify users about pending Jan CPU release

jerboaa:
Draft pull request for Jan 2020 security release.

PR #667, opened 2020-01-23
Revert "Add banner to notify users about pending Jan CPU release"

smlambert:
Reverts AdoptOpenJDK/openjdk-website#657


jerboaa:
@smlambert Next time perhaps not use `git revert` but use a new commit which just comments it out.

PR #683, opened 2020-03-04
Use $releasever and $basearch in Centos example

makubi:
### Checklist

### DIFF
--- a/src/handlebars/installation.handlebars
+++ b/src/handlebars/installation.handlebars
@@ -345,15 +345,14 @@
        </p>
 
        <p><strong>RPM installation on Centos, RHEL, or Fedora</strong></p>
-       <p>The following steps describe how to install an RPM package for Centos 7. To install an RPM for RHEL or Fedora, or
-         to install a different version of Centos, switch the <tt>baseurl</tt> value.
+       <p>The following steps describe how to install an RPM package for Centos. To install an RPM for RHEL or Fedora update the <tt>baseurl</tt> value accordingly.
        </p>
        <p>1. Add the appropriate RPM repository to your <tt>/etc/yum.repos.d/adoptopenjdk.repo</tt> file, by running the following command:</p>
        <pre><code class="highlighted">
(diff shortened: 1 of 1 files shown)


aahlenst:
I copied and pasted

```
cat <<EOF > /etc/yum.repos.d/adoptopenjdk.repo
[AdoptOpenJDK]
name=AdoptOpenJDK
baseurl=http://adoptopenjdk.jfrog.io/adoptopenjdk/rpm/centos/$releasever/$basearch
enabled=1
gpgcheck=1
gpgkey=https://adoptopenjdk.jfrog.io/adoptopenjdk/api/gpg/key/public
EOF
```

into a running CentOS 8 container and it resulted in

```
[AdoptOpenJDK]
name=AdoptOpenJDK
baseurl=http://adoptopenjdk.jfrog.io/adoptopenjdk/rpm/centos//
enabled=1
gpgcheck=1
gpgkey=https://adoptopenjdk.jfrog.io/adoptopenjdk/api/gpg/key/public
```

Any special prerequisites needed?

aahlenst:
@karianna Does it work for you?

makubi:
@aahlenst You are right, sorry. I did not test the copy & paste. What do you think about adding `\`s before the `$`? I will open another PR.

aahlenst:
@makubi Could you elaborate what you're trying to achieve? Does for example CentOS define those variables in the environment so that they are automatically substituted with the correct values? Or should `$releasever` denote a placeholder that the user is expected to replace by hand? If the latter, why include `$` at all and it needs an explanation in the surrounding text.

makubi:
@aahlenst Redhat defines those variables, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-using_yum_variables
I am trying to achieve a version-agnostic configuration. Other repositories like CentOS-Base use these variables as well.

aahlenst:
Great. So please do whatever is needed to make the commands copy & paste compatible and I'll happily merge.

PR #700, opened 2020-04-08
allow website to show ea builds

gdams:
##### Checklist

- [x] `npm test` passes

This PR adds an early access tag:

```js
  if ((version.pre) && (version.pre == 'ea')) {
    return true
  } else { 
    return false
  }
```
Sample:
Screenshot 2020-04-08 at 11 36 29

### DIFF
--- a/src/handlebars/releases.handlebars
+++ b/src/handlebars/releases.handlebars
@@ -64,7 +64,11 @@
                     <tr>
                        <td class="main-download__variant__name blue-bg">
                           <span>\{{release_name}}</span>
+                          \{{#if early_access}}
+                          <span class="early-access">AdoptOpenJDK</span>
+                          \{{else}}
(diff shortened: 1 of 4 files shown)


lumpfish:
For the immediate requirement (AArch64 jdk11 openj9) must understanding is that the build pipeline will be submitted specifying 'release' (which I presume would make pre=ga).

lumpfish:
Regarding the "pre" tag -  it is created when the "-ea" string is found in the java -version output, so a GA build will not carry the tag at all. So checking for pre=ea is all that is necessary in the web pages for tagging the builds (openjdk_version need not be checked and pre=ga is impossible)

karianna:
@gdams It's hard to see evidence on staging that this worked - no binaries around to pick up?

gdams:
@karianna that's correct, we will have to wait for the EA release to happen in order to see this

PR #730, opened 2020-06-03
add black lives matter banner

gdams:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/scss/styles-1-large-main.scss
+++ b/src/scss/styles-1-large-main.scss
@@ -671,7 +671,7 @@ footer p {
    color: black;
 }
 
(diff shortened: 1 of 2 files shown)


src/scss/styles-1-large-main.scss
```diff
@@ -671,3 +671,9 @@ footer p {
    color: black;
 }
 
+.css-1y7sxf6, [data-css-1y7sxf6] {
```
hendrikebbers, line 674: can we define a better name for the classes? Or is this any name convention I do not know about?

gdams: fixed

PR #745, opened 2020-07-06
Runtime version support for IcedTea-Web

hendrikebbers:
Adds AdoptOpenJDK / OpenJDK support information (supported versions) to the `IcedTea-Web `and `Migration` pages.


hendrikebbers:
@karianna that is not a new page. I only added a section but it looks like my commit changed the line feed of the file. Will have a look.

karianna:
@hendrikebbers - Did you manage to check what you needed?

hendrikebbers:
@gdams I think the problem is on my local machine. From time to time git changes the LF of files that I commit from LF to CRLF. Have not found a solution for that so far... Changed the linefeed by hand. Now it looks good here.

PR #783, opened 2020-09-07
Add RHEL/CentOS8 as supported platforms

sxa:
Signed-off-by: Stewart Addison <sxa@uk.ibm.com>
- [x] `npm test` passes

CentOS8 is not yet available on s390x but this will add RHEL8. @smlambert I explicitly want your confirmation that you're happy with this before adding it given that we don't have systems running this level across ll platforms at present.


gdams (approved):

smlambert (commented):
for posterity, approved 👍

PR #784, opened 2020-09-07
Add arm32 Ubuntu distributions as supported

sxa:
At the moment the supported platforms pages have nothing listed for arm32/armv7l/aarch32

Signed-off-by: Stewart Addison <sxa@uk.ibm.com>

##### Checklist

- [ ] `npm test` passes

### DIFF
--- a/src/handlebars/supported_platforms.handlebars
+++ b/src/handlebars/supported_platforms.handlebars
@@ -45,6 +45,7 @@
                     <th>x64</th>
                     <th>ppc64le</th>
                     <th>s390x</th>
+                    <th>aarch32</th>
                     <th>aarch64</th>
                 </tr>
             </thead>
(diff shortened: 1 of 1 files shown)


src/handlebars/supported_platforms.handlebars
```diff
@@ -45,6 +45,7 @@
                     <th>x64</th>
                     <th>ppc64le</th>
                     <th>s390x</th>
+                    <th>aarch32</th>
```
smlambert, line 48: Does the API serve up as aarch32 or as arm32?  I have seen both ways of naming this platform... we should likely attempt to consolidate on 1 way of naming...

smlambert (approved):
Other than the question around naming aarch32 versus arm32, the changes LGTM.  Will leave it up to @sxa to merge (depending on what you decide to do with the name).

gdams (approved):

src/handlebars/supported_platforms.handlebars
sxa: Neither ... The API uses `arm` at the moment for 32-bit ARM but I agree this was the one thing I was undecided about, but chose to make it the same as the other columns that were in the file for now.

sxa:
General approval that this is a tactical fix and we should look at merging naming outside of this PR

PR #798, opened 2020-10-06
Clarify refreshing package list and installing deb

karth295:
1) Move `apt-get update` into the code section. It was a bit hidden in the description before.
2) Remove brackets around package to install, as it prevents simply copying and pasting the command.

#### Checklist
- [x] documentation is changed or added (if applicable)


src/handlebars/installation.handlebars
```diff
@@ -334,9 +334,12 @@
       <pre><code class="highlighted">sudo apt-get install -y software-properties-common</code></pre>
 
       <p>Then repeat the first command.</p>
-      <p>3. Refresh your package list with <tt>sudo apt-get update</tt> and then install your chosen AdoptOpenJDK package. For example, to
+      <p>3. Refresh your package list with and then install your chosen AdoptOpenJDK package. For example, to
```
karianna, line 337: Grammar doesn't make sense here?  Refresh your package list with and...

karianna (changes requested):

gdams:
@aahlenst do you think we need to clarify here? To me, the original wording seems pretty clear

aahlenst:
Well, `apt-get update` was moved to the code sample. Maybe it's more visible that way. But I don't know if it really helps. The problem is the entire page.

What we really need is a nice assistant that helps you generate the right configuration and commands but that's out of scope here.

gdams:
okay, let's close this PR for now and we can revisit

src/handlebars/installation.handlebars
karth295: Oops, yeah my PR should have deleted the word `with` as well.

PR #817, opened 2020-10-21
update supported version for j9 15+

gdams:
##### Checklist

- [x] `npm test` passes

### DIFF
--- a/src/js/common.js
+++ b/src/js/common.js
@@ -63,6 +63,7 @@ module.exports.getBinaryExt = (searchableName) => lookup[searchableName].binaryE
 module.exports.getInstallerExt = (searchableName) => lookup[searchableName].installerExtension;
 
 // gets the Supported Version WITH PATH when you pass in 'searchableName'
+// Version numbers use >= logic and need to be specified in ascending order
 module.exports.getSupportedVersion = (searchableName) => {
   let supported_version = lookup[searchableName].supported_version;
   if (typeof supported_version === 'object') {
--- a/src/js/common.js
+++ b/src/js/common.js
@@ -63,7 +63,23 @@ module.exports.getBinaryExt = (searchableName) => lookup[searchableName].binaryE
 module.exports.getInstallerExt = (searchableName) => lookup[searchableName].installerExtension;
 
 // gets the Supported Version WITH PATH when you pass in 'searchableName'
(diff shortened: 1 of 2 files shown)


sxa (changes requested):
While this appears to work, I think this needs some documentation/comments added to make it clear why `11` isn't included in the list along with `8` and `15` otherwise it's rather non-obvious to someone looking at this why and the reason is somewhat hidden in the code in `getSupportedVersion` so _at the very least_ it needs a sentence above the function indicating how the string is selected.

src/js/common.js
```diff
@@ -63,7 +63,23 @@ module.exports.getBinaryExt = (searchableName) => lookup[searchableName].binaryE
 module.exports.getInstallerExt = (searchableName) => lookup[searchableName].installerExtension;
 
 // gets the Supported Version WITH PATH when you pass in 'searchableName'
-module.exports.getSupportedVersion = (searchableName) => lookup[searchableName].supported_version;
+module.exports.getSupportedVersion = (searchableName) => {
+  let supported_version = lookup[searchableName].supported_version;
+  if (typeof supported_version === 'object') {
```
karianna, line 68: `supported_version === 'object'` - This is why I weep at dynamic typing ;-).  V2 of the website - I vote we go Typescript

gdams: agreed :D

gdams:
> While this appears to work, I think this needs some documentation/comments added to make it clear why 11 isn't included in the list along with 8 and 15 otherwise it's rather non-obvious to someone looking at this why and the reason is somewhat hidden in the code in getSupportedVersion so at the very least it needs a sentence above the function indicating how the string is selected.

@sxa updated, comments in JSON are always horrible but it's the best that I can do

sxa:
> @sxa updated, comments in JSON are always horrible but it's the best that I can do

I was more thinking about commenting the js function to be honest, although certainly a comment in there is welcome (even though saying it's >= isn't entirely accurate since 16>=8, but any clearer wording would be longer unless we could link into the parsing function)

sxa:
Also (I don't know why this randomly came into my head at 5AM) is there a "booboy-trap" here for anyone updating or trying to use this new logic whereby if the JVM version numbers are not in ascending order it won't work?

sxa (approved):
Thanks - we should consider if we can make the logic a bit more maintainable/reliable to avoid people accidentially introducing errors in the JSON in the future though

PR #824, opened 2020-10-26
Updates to support page for next LTS, release dates and Dragonwell

sxa:
Miscellaneous updates to the doc which I'd spotted while doing the quarterly updates over the weekend in my last PR :-)

Signed-off-by: Stewart X Addison <sxa@redhat.com>

##### Checklist

- [x] `npm test` passes
- [x] documentation is changed or added (if applicable)


src/handlebars/support.handlebars
```diff
@@ -56,127 +56,134 @@
...
         <a href="https://www.eclipse.org/openj9/" target="_blank">Eclipse OpenJ9</a>,
         <a href="https://aws.amazon.com/corretto/" target="_blank">Amazon Corretto</a>,
-        and <a href="https://sap.github.io/SapMachine/" target="_blank">SAP Machine</a>.
+        <a href="https://sap.github.io/SapMachine/" target="_blank">SAP Machine</a>
+        and <a href="https://github.com/alibaba/dragonwell11" target="_blank">Alibaba Dragonwell</a>.
       </p>
       <p>
-        OpenJDK and Eclipse OpenJ9 provide a new feature release every six months, and a maintenance/security update based upon each active release every three months.
+        OpenJDK and Eclipse OpenJ9 provide a new feature release every six months, and a maintenance/security update based upon each
+        active release every three months. The release dates for those from the OpenJDK project are the
```
tellison, line 64: Consider (adds "usually"):
 "The release dates for those from the OpenJDK project are the" -> "The release dates for those from the OpenJDK project are usually the"
or point people to the definitive source of info.

src/handlebars/support.handlebars
```diff
@@ -56,127 +56,134 @@
...
+        17th</a> of January, April, July and October.
         We will follow this schedule for publishing binary releases from AdoptOpenJDK to ensure you get the latest, most secure builds.
       </p>
       <p>
         In addition, every three years one feature release will be designated as a Long Term Supported (LTS) release.
         We will produce LTS releases for at least four years. This assurance will allow you to stay on a well-defined code stream,
         and give you time to migrate to the next, new, stable, LTS release when it becomes available.
       </p>
-      <p>Based upon this roadmap, and starting with Java 8:</p>
+      <p>Based upon this roadmap, and starting with Java 8 (Currently supported releases in bold)
```
tellison, line 74: Consider (case):
"Currently" -> "currently"

src/handlebars/support.handlebars
```diff
@@ -56,127 +56,134 @@
...
+      here is the release timetable for the various releases:</p>
```
tellison, line 75: Maybe drop the first "release" in that sentence?

src/handlebars/support.handlebars
```diff
@@ -56,127 +56,134 @@
...
             <strong>jdk-15.0.1+9<br/>
-                20th October 2020</strong>
+                20th Oct 2020</strong>
           </td>
           <td>
             <strong>jdk-15.0.1<br/>
-              19th January 2021</strong></td>
-          <td>March 2021</td>
+              19th Jan 2021</strong></td>
+          <td>Mar 2021</td>
```
tellison, line 163: Missing a `strong` tag

src/handlebars/support.handlebars
sxa: I may be misinterpreting you, but the Oracle link says `They are released on the Tuesday closest to the 17th day of January, April, July and October.` and I've replicated that here - do you not consider that sufficiently definitive?

tellison: I meant it may be safer to **refer** to that definitive source rather than **replicate** it, such that if OpenJDK update their text you won't need to notice their change and make a corresponding update the AdoptOpenJDK website.

sxa: Hmmm I see what you're saying now, although making the text with the dates a hyperlink makes it about as clear as it can be where the definitive source is and as an end-user it would be a pain to be forced to navigate away from the site to see it while looking at the table. If it was something that I felt was likely to change I'd agree, but I do think it would be better to have that information explicit next to the table in this case. Perhaps a "Subject to change, check the link" disclaimer against it?

gdams:
@sxa is this good to go now?

karianna (approved):
LGTM

PR #836, opened 2020-11-06
actions: add production action

gdams:
##### Checklist

- [ ] `npm test` passes

### DIFF
--- a/.github/workflows/deploy-to-gh-pages.yml
+++ b/.github/workflows/deploy-to-gh-pages.yml
@@ -1,0 +1,52 @@
+name: Deploy to GitHub Pages
+on:
+  push:
+    branches: [ master ]
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        ref: master
+        path: website
+      
+    - uses: actions/setup-node@v1
+      with:
+        node-version: '10'
+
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-to-gh-pages.yml
```diff
@@ -0,0 +1,52 @@
...
+        npm install
+        gulp build
+
+    - name: Create Directory Stucture
+      run: |
+        mkdir tmp
+        mv website/sitemap.xml tmp/
+        cd website
+        find . -type f ! -name '*.html' ! -name '*.pdf' -maxdepth 1 -mindepth 1  -delete
+        find . -type d -not -name 'dist' -not -name '.git' -maxdepth 1 -mindepth 1 -exec rm -rf {} \;
```
aahlenst, line 37: What's the purpose of those two?

aahlenst (commented):
General question: Why is "Create Directory structure" part of CI and not gulp? If it was part of the gulp build, I could try that locally?

.github/workflows/deploy-to-gh-pages.yml
gdams: in order to deploy to github pages we need to remove the development files. Gulp doesn't do this because it would be impossible to develop at the same time.

I'm happy to look at optimisations in the future but for now I'm just using the scripts from the jenkins job

aahlenst (approved):

PR #855, opened 2020-11-26
Add ojdkbuild as possible OpenJFX source

aahlenst:
According to https://github.com/AdoptOpenJDK/openjdk-build/issues/577#issuecomment-732978013, ojdkbuild is a viable source for OpenJFX 8 builds. This PR adds it to the OpenJFX FAQ.

cc @jschwartzenberg


karianna (approved):
lgtm

jschwartzenberg:
Thanks for this! It might better fit one question higher though as that one is about version 8.

aahlenst:
@jschwartzenberg You're totally right. Fix coming with #856.

PR #1081, opened 2021-09-23
Debian repository installation: Fix path to keyring in chown

joanbm:
The chown command in #1080 is obviously wrong, as it should be applied to the path of the file after moving, not the one before moving.

##### Checklist
- [X] `npm test` passes
- [X] documentation is changed or added (if applicable)
- [X] contribution guidelines followed [here](https://github.com/AdoptOpenJDK/openjdk-website/blob/master/CONTRIBUTING.md)

### DIFF
--- a/src/handlebars/installation.handlebars
+++ b/src/handlebars/installation.handlebars
@@ -369,7 +369,7 @@ rm adoptopenjdk-keyring.gpg
         <li> Save keyring in root directory (Create the keyrings directory)
           <pre>
             <code class="highlighted">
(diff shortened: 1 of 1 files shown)


joanbm:
PS: ...this is embarrassing. I swear I built the website from a fresh copy twice to make sure I didn't make a stupid mistake before sending the PR, but somehow I still changed the order of the commands at the last moment and forgot to test it.

karianna (approved):
LGTM

PR #1523, opened 2025-05-13
Move to vite based build system to be able to keep site up to date with security vuls

karianna:


gdams:
nice refactor! I see two issues though:

1) The release assets aren't loading on the page: https://deploy-preview-1523--vibrant-hugle-19d716.netlify.app/archive
2) Images don't seem to be rendering for me, is it the same on your side? I think for this one you need to move all the images in `src/assets` into the `public` folder.

public/vite.svg
gdams, line 1: guessing this isn't needed

src/javascript.svg
gdams, line 1: guessing this isn't needed

karianna:
> nice refactor! I see two issues though:
> 
> 1. The release assets aren't loading on the page: https://deploy-preview-1523--vibrant-hugle-19d716.netlify.app/archive
> 2. Images don't seem to be rendering for me, is it the same on your side? I think for this one you need to move all the images in `src/assets` into the `public` folder.

Yeah oddly these both work fine for me locally when running npm run dev, so I suspect a subtle deploy difference there.

karianna:
@gdams - I think this can go another review now.  There's some CORS stuff which can stop the Adoptium API from responding but that's a local dev thing which you can work around.  Would be good to re-trigger netlify but not sure how to do that just yet!

gdams (approved):

